Issue #151, opened 2015-09-26
Markdown syntax help


dmarrazzo:
It could be useful especially for newbies to have a toolbar with all markdown formatting codes.
E.g. Titles, bold, italic, list items, images, links, etc.
A good example would be the stackoverflow text box.

mitya57:
I won't add the new toolbar (as I try to keep the design minimal), but I agree that HTML tags box should better be replaced with something more Markdown-ish.

dmarrazzo:
If you agree, I would try to change the tags menu with Markdown syntax.
I will work on my fork then I'll propose the pull.

mitya57:
I have now merged @dmarrazzo's changes to replace the Tags box with a Formatting box, so I think this can be closed now.

PR #153, opened 2015-09-29
Tags helper changed from html to md

dmarrazzo:
it addresses issue #151


ReText/window.py
```diff
@@ -697,7 +697,7 @@ def getHtml(self, includeStyleSheet=True, includeTitle=True,
 			headers += '<link rel="stylesheet" type="text/css" href="%s">\n' \
 			% cssFileName
 		if includeMeta:
-			headers += ('<meta name="generator" content="ReText %s">\n' %
+			headers += ('<meta name="generator" content="ReText %s" >\n' %
```
mitya57: Why the space?

ReText/window.py
```diff
@@ -211,8 +211,8 @@ def __init__(self, parent=None):
 			trig=lambda: self.insertChars('*'))
 		self.actionUnderline = self.act(self.tr('Underline'), shct=QKeySequence.Underline,
 			trig=lambda: self.insertTag('u'))
-		self.usefulTags = ('a', 'big', 'center', 'img', 's', 'small', 'span',
-			'table', 'td', 'tr', 'u')
+		self.usefulTags = ('header', 'italic', 'bold', 'numbering', 'bullets', 'image', 'link',
```
mitya57: There are already bold/italics/underline actions in Edit menu, please don't duplicate the functionality.

ReText/window.py
dmarrazzo: It was unintentional.

ReText/window.py
dmarrazzo: What do you think to place everything in one place to have a consistent behaviour?
I see two options:
 - remove the Formatting submenu (I would prefer because the toolbar is more accessible)
 - remove the Tags tool and move in Formatting

mitya57: The toolbar is indeed more accessible, but the menu actions also provide shortcuts. Don't know if someone uses them, though. So maybe your suggestion makes sense.

dmarrazzo: it would be perfect to save even the key shortcuts.

ReText/window.py
```diff
@@ -1161,23 +1161,48 @@ def insertChars(self, chars):
...
+			selectedText = tc.selectedText()
+			if not selectedText:
+				selectedText = 'alt text'
+			toinsert = ('['+ selectedText +'](url)')
+		elif ut == 'code':
+			selectedText = tc.selectedText()
+			if not selectedText:
+				toinsert = '\n\n\t'
+			else: 
+				toinsert = ('`'+ selectedText +'`')
```
dmarrazzo: Please note that `code` command has two behaviour depending on whether or not text is selected.
I'm not sure about that:
 - it's comfortable having two result with one command
 - it could be confusing for the newbie

mitya57: Can we have two separate commands: "inline code" and "code block"?

ReText/window.py
```diff
@@ -1161,23 +1161,48 @@ def insertChars(self, chars):
...
+			toinsert = ('!['+ selectedText +'](url)')
+		elif ut == 'link':
+			selectedText = tc.selectedText()
+			if not selectedText:
+				selectedText = 'alt text'
+			toinsert = ('['+ selectedText +'](url)')
+		elif ut == 'code':
+			selectedText = tc.selectedText()
+			if not selectedText:
+				toinsert = '\n\n\t'
```
mitya57: Please use four spaces instead of `\t` (in inserted text), to follow the Markdown standard.

ReText/window.py
mitya57: Do you mean you want both actions and combobox items?

ReText/window.py
dmarrazzo: Done

ReText/window.py
dmarrazzo: Done

ReText/window.py
dmarrazzo: yes, combobox + key shortcut.
Maybe the problem is make them clear for users, I think that the shortcut should be declared in a help page.

dmarrazzo: I'm trying to add a "stand alone" shortcut without lucky!
I found this example that should work with Qt4

    self.boldShortcut = QShortcut(QKeySequence.Bold, self, None, self.insertTag("bold"))

But it does not work in Qt5! Any idea?

mitya57: I don't know why it wouldn't work.

Keeping both actions and toolbar items is also fine, just maybe we can make them use the shared code?

My priority now is fixing an XSettings crash on Xfce and getting a bug fix release out, after that I will merge your pull request.

mitya57:
Hi, sorry for the delay, my schedule was busy and I was not able to look at this.

* I don't think you need to insert newlines when starting a list / code block / etc. I think the user has already started a new paragraph if (s)he clicks on that. It is quite confusing if one i.e. creates a new document, clicks on the helper item, and gets lots of newlines.
* When you start bold italics or underline, the cursor is inside the created block, i.e. `**>cursor<**`. However, when starting image or link, the cursor appears after it. Is it possible to put the cursor inside the newly created block, i.e. `[>cursor<](url)`?

Re the shortcuts, let's leave the current QAction's for bold/italics/underline in place, just make sure they work correctly and share code with the helper items.

Also, there are too many commits in your pull request, please squash them into a single one.

ReText/window.py
```diff
@@ -39,7 +39,7 @@
 from PyQt5.QtWidgets import QAction, QActionGroup, QApplication, QCheckBox, \
  QComboBox, QDesktopWidget, QDialog, QFileDialog, QFontDialog, QInputDialog, \
  QLineEdit, QMainWindow, QMenuBar, QMessageBox, QSplitter, QTabWidget, \
- QTextBrowser, QTextEdit, QToolBar
+ QTextBrowser, QTextEdit, QToolBar, QShortcut
```
mitya57: Please keep the list sorted. `QShortcut` should go between `QMessageBox` and `QSplitter`.

mitya57:
@dmarrazzo Ping? If you fix the issues I mentioned [above](https://github.com/retext-project/retext/pull/153#issuecomment-154742157), I will merge it.

dmarrazzo:
@mitya57 sorry for the delay

dmarrazzo:
hello @mitya57, can you merge my code? I'm pretty sure that other users will enjoy it? Thank you

mitya57:
@dmarrazzo Sorry for the delay! I now squashed your commits into one big, resolved the conflict, and refactored the code a bit.

dmarrazzo:
Ok thank you

PR #165, opened 2015-11-07
Add *.bat and change icons for Windows users

Prodesire:
1.Add *.bat scripts for  Windows users to lrelease *.ts to *.qm, and remove *.qm in locale directory. The run.bat will firstly excute .\locale\lrelease_retext.bat to check if there is *.ts which doesn't have the relevant *.qm file, otherwise, it will lrelease *.ts to *.qm. Then, run.bat will excute python script to start the ReText.
2. Change the icons for Windows.


mitya57:
Hi, thanks for caring about ReText. However, this pull request in its current form is a mess, so I will not merge it.

* I do not care what `.idea` is, please do not add that to Git.
* Please do not add Chinese text to README. It makes it unreadable and I don't even know what you have written there :)
* Please do not change tabs indentation to spaces. I know that PEP-8 recommends spaces, but we are using tabs for historical reasons and changing it to spaces will make things like `git blame` not work.
* Please do not commit icons to Git. I am maintaining a separate icons pack for Windows and OS X users, but it should not be in the default repository. The icons pack is currently using the Oxygen icon theme from KDE SC 4. If you want to suggest any alternative icon themes, please let me know. Those themes should be free software, of course.

Now, the description of this pull request mentions “`*.bat` scripts for Windows users to lrelease `*.ts` to `*.qm`”. However, I do not understand this. Why can't one just use the `lrelease locale/*.ts` command? If that is not working on Windows, can I somehow fix the `build_translations` command in `setup.py` to make it working?

That said, I will consider **clean** pull requests if they fix existing code to make it work on Windows, but I would not like new scripts or files to be added, unless it is critical to have them to support Windows.

Prodesire:
Hello, mitya57! I am very thankful for your reply. I have to admit that the pull request is in a very considerable mess. Actually I just wanted to pull the last commit about converting scripts. For some reason( I am not so familiar with pull request in github, sorry about that), the result seems not like what I think.
I have tried  the `lrelease locale/*.ts` command in Windows cmd, but it not worked. So, I wrote some scripts to do the same thing. In Windows platform, using `python setup.py install` will not help users to use retext, because the `retext` script installed in `Python34(may be other versions)\Scripts\`  doesn't work properly in Windows. So, for the users on Windows who don't run `python setup.py install` command, using `run.bat`, `lrelease_retext.bat` and `lrelease_clear.bat` may be helpful to run ReText, convert `*.ts` to `*.qm`, and clear `*qm`.
As far as I am concorned,  if running `python setup.py install` command in ReText root dictory could generate retext.bat script, which could be run  on Windows and used for running ReText, in `Python34(may be other versions)\Scripts\`, will be very helpful for Windows users. Thanks for your reading :) .

mitya57:
Can you explain why `lrelease locale/*.ts` is not working? Does it work on individual files (i.e. `lrelease locale/lang.ts` for every `lang`)? Does it require `.qm` files to be removed before running?

I would very much like to add support for this to `setup.py`, instead of shipping `.bat` files which I do not understand.

I should also add that the release tarballs (i.e. on PyPI) already **do** contain `.qm` files, and for those running the Git snapshot, it's normal to require to run `setup.py build` (not install).

Prodesire:
I have clear all `*.qm` files, and just open cmd in root directory of ReText to run `lrelease .\locale\*.ts` (Well, this form in Windows cmd is a little bit different from in Linux). The not working result maybe due to no regular support of  `lrelease` command in Windows, so I have to run command like `lrelease .\locale\retext_ca.ts` to build every single `.ts` file to `.qm`. That's why I have to write the .bat.

I think add release tarballs which contain `.qm` files is good for us to use ReText more easily. Hope for that comming soon!

Prodesire:
By the way, it is more convenient if add `live preview` button to the ReText tool bar. As for me, I would like to see final result of .md of .rst text on real-time. Every time I run ReText and have to enable `live preview` function through the menu, makes me feel a little bit unconvenient. I wonder if you can add this feature to next version, and if possible, remember the last mode the user set before close ReText, and  restore it when open ReText next time.

mitya57:
The `.qm` files are **already** in the tarballs on PyPI, no need to wait :)

If I implement calling commands similar to `delete .\locale\retext_lang.qm && lrelease .\locale\retext_lang.ts` during `setup.py build`, will that work for you? It's quite easy to do it in Python with `os.path.join`, `os.unlink` and friends.

Re the live preview action: does <kbd>Ctrl</kbd>+<kbd>L</kbd> not work?

Prodesire:
Well, I will have a look at PyPI to find out `.qm` files.

Implementing commands you said during `setup.py build` is a good idea and of course it will work for me. Using Python to do such work is much easier and more readable than batch script in Windows.

`ctrl + L` works well. I think we should regard `preview` and `live preview` as the same class for they have similar function. So putting one in the tool bar means the other should also be putted. What do you think about this? ^_^

mitya57:
I will look at putting live preview to the toolbar.

Actually, can you check what happens if you run `setup.py build` now? What does it print after the "running build_translations" line?
I suspect that the current code could already work correctly.

mitya57:
I have pushed a commit that makes the Live Preview action accessible from the toolbar: [action-live-preview](https://github.com/retext-project/retext/tree/action-live-preview). Can you check if it suits your needs?

Prodesire:
I so sorry that I haven't run `setup.py build` before. This command works well and after running it, I could see .qm files. The thing is when I firstly used ReText, I run `setup.py install` and I couldn't find .qm files (which seems obviously now). This kind of experience makes `install` command haunt me, so I mess up with `install` and `build`. Now, I would confirmedly say `setup.py build` doesn't have any problem!

I have downloaded the lastest version of ReText, but I can't find `live preview` button (or maybe else) in toobar. I don't know if I misunderstand your commit or maybe there is some trouble on Windows?

Prodesire:
Well, I find the reason why I miss the `live preview` feature. I downloaded the project from the main project of retext, but your commit is in the tree of it.

Now, I have seen the `live preview` button as the child of `preview` in toolbar. Thanks for doing that!

mitya57:
Ok, I have merged that commit into master now.

Prodesire:
Great! I saw the changes.

PR #166, opened 2015-11-19
internal links

Tamriel:
When links like [test](test) are clicked, the file test.md is opened.
It has to be located next to the current opened file.
Relative pathes like [test](../test) or [test](folder/test) are also possible.

### DIFF
--- a/ReText/tab.py
+++ b/ReText/tab.py
@@ -72,9 +72,7 @@ class ReTextTab(QObject):
 	def createPreviewBox(self):
 		if globalSettings.useWebKit:
 			return self.createWebView()
-		browser = QTextBrowser()
-		# TODO: honor globalSettings.handleWebLinks?
-		browser.setOpenExternalLinks(True)
+		browser = ReTextPreview(self)
 		return browser
 
 	def getSplitter(self):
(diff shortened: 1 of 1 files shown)


mitya57:
Thanks for your pull request. I have made some comments on your commit, please fix that or let me know if I should do that myself. I would prefer if you amended the commit instead of adding a new one.

(I should really have left these comments on the PR diff, not on the commit, but too late…)

Tamriel:
Thanks for your insightful feedback!

    if link.url().startswith('http'):
        QDesktopServices.openUrl(link)

> Can this really happen? We call `setOpenExternalLinks(True) ` which according to the documentation means: 
>> QTextBrowser should automatically open links to external sources using QDesktopServices::openUrl() instead of emitting the anchorClicked signal.

I `self.setOpenLinks(False)`, because if set to True, links to other files will unsuccesfully be opened as anchors. Therefore the distinction of cases is necessary.

ReText/tab.py
```diff
@@ -188,3 +189,30 @@ def installFakeVimHandler(self):
...
+	"""
+
+	def __init__(self, tab):
+		QTextBrowser.__init__(self)
+		self.tab = tab
+		self.setOpenLinks(False)  # if set to True, links to other files will unsuccesfully be opened as anchors
+		self.anchorClicked.connect(self.openInternal)
+
+	def openInternal(self, link):
+		if link.url().startswith('http'):
```
mitya57: How about using `not link.url().isRelative()` instead? It is better because:

* It will fail on different protocols, not only http (i.e. ftp, torrent).
* It will not fail on files named i.e. http.mkd :)

ReText/tab.py
```diff
@@ -188,3 +189,30 @@ def installFakeVimHandler(self):
...
+			QDesktopServices.openUrl(link)
+		else:
+			currentFileInfo = QFileInfo(self.tab.fileName)
+			current_file_dir = currentFileInfo.absoluteDir().absolutePath()
+			if re.search('\.\w*$', link.url()):  # file suffix exists in link
```
mitya57: This also matches a dot in a directory name, doesn't it? Maybe call `toLocalFile()` first, and then check if it has a suffix?

ReText/tab.py
```diff
@@ -188,3 +189,30 @@ def installFakeVimHandler(self):
...
+				suffix = ''
+			else:
+				suffix = '.' + currentFileInfo.completeSuffix()
+			file_to_open = current_file_dir + '/' + link.url() + suffix
```
mitya57: Err, no hardcoded '/', please. Please use `QDir.filePath` method instead.

ReText/tab.py
```diff
@@ -188,3 +189,30 @@ def installFakeVimHandler(self):
...
+	"""
+	When links like [test](test) are clicked, the file test.md is opened.
+	It has to be located next to the current opened file.
+	Relative pathes like [test](../test) or [test](folder/test) are also possible.
+	"""
+
+	def __init__(self, tab):
+		QTextBrowser.__init__(self)
+		self.tab = tab
+		self.setOpenLinks(False)  # if set to True, links to other files will unsuccesfully be opened as anchors
```
mitya57: If you call `setOpenLinks(False)`, then does it make sense to use `setOpenExternalLinks(True)` above? (Ideally it should honor `globalSettings.handleWebLinks` option like the WebKit backend does, but I can take care about that myself.)

Also, please cut the long line.

mitya57:
Much better, thanks, though still some issues :)

Tamriel:
Beautiful code now, just like yours ;)
Do you use 80 or 100 columns wide code? I prefer 100.

mitya57:
I try to use 100, but sometimes the code gets even wider… So new code should not use lines longer than 100 symbols.

mitya57:
Finally I got around to your code. I have fixed it to handle some non-standard cases (HTML files, files unknown to ReText, files with absolute paths, anchor links), and merged. Thanks!

PR #169, opened 2015-11-30
desktop file: Use executable name instead of full path

### LINKED ISSUE #168: Failed to launch ReText if installation prefix isn't /usr
dolphin_002
Remove the full path from the desktop entry's `Exec` key solve the issue.

Lin-Buo-Ren:
Application will fail to be launched if the installation prefix isn't
/usr, fixes #168 .

Signed-off-by: 林博仁(Buo-Ren Lin) <Buo.Ren.Lin@gmail.com>


mitya57:
Merged, thanks!

PR #172, opened 2015-12-20
Added setting 'Automatically open live preview for new documents'

Tamriel:


mitya57:
Sorry, I won't merge all this bulk of commits :)

Commenting on the first commit (which corresponds to the title of this PR): the option is misleading. For me, the "new document" is what happens when one selects "File → New document" (i.e. the empty document). However your commit is opposite — it does not affect new documents and affects only opening files.

Perhaps you can do that in ReTextTab constructor, and call the option somehow different?

PR #180, opened 2016-01-18
Made the doc type update when the markup is changed

Potato42:
When you change the default markup type, it should change things like making the tags and symbols enabled/disabled, etc.  Currently it does not, but this pull request seems to fix that.

In terms of testing, I ran the program and changed the markup type and it seemed to work.


mitya57:
Fixed it a bit differently. Thanks @Potato42!

PR #187, opened 2016-01-20
fixing webkit preview color on dark qt theme

farseerfc:
When using a dark qt theme (e.g. breeze-dark), webkit preview have a
dark text color on a dark background, which is hardly visible.
This fix adds a CSS that pulls text color from default QtPalette.

### DIFF
--- a/ReText/tab.py
+++ b/ReText/tab.py
@@ -27,7 +27,7 @@ except ImportError:
 	ReTextFakeVimHandler = None
 
 from PyQt5.QtCore import Qt, QDir, QFile, QFileInfo, QObject, QTextStream, QTimer, QUrl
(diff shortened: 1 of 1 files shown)


farseerfc:
This is a screenshot with the fixed preview(above) and original behavior(below).
screenshot_20160121_053424

mitya57:
@farseerfc Thanks for the pull request — I fixed it in a bit different place. Please check if the fix still works for you.

farseerfc:
@mitya57 Yes. Tested d5e49a779f346e9e8f0895ab74d59674a5866a6d. This fixed my problem. Thank you!

mitya57:
Ok, thanks for testing!

PR #189, opened 2016-01-31
glob locale files lazy during setup

farseerfc:
Currently `setup.py build` will generate locale translations,
and `setup.py install` will install these generated translations.
However `setup.py install` from a clean cloned repo will not install these
translations because they are generated **after** glob could find these
files. This fix use a generator to lazy glob these translations after they are generated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,9 @@ def build_translations():
 	if error:
 		print('Failed to build translations:', error)
 
+def glob_lazy(path):
+    yield from glob(path)
+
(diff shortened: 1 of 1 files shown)


mitya57:
Thanks! There is an [iglob](https://docs.python.org/3/library/glob.html#glob.iglob) function in glob module so I used that instead.

farseerfc:
Wow! Didn't know iglob will also do the trick. Great!

PR #190, opened 2016-01-31
QT_SELECT=5 instead of QT_SELECT=qt5

farseerfc:
The `setup.py` calls `lrelease` with `QT_SELECT=qt5` set.
On Arch Linux this will error with:
> lrelease: could not find a Qt installation of 'qt5'

The output of `qtchooser -list-versions` on Arch Linux is:
> 4
> 5
> default

This is the same for
[LFS](http://www.linuxfromscratch.org/blfs/view/7.4/general/qtchooser.html)

This fix changed "qt5" to "5".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ def build_translations():
 	error = None
 	for ts_file in glob(join('locale', '*.ts')):
 		try:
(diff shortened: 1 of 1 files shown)


mitya57:
Thanks! I was initially writing this on Debian where both `qt5` and `5` exist.

PR #209, opened 2016-05-17
Improve Windows compatibility

JGoutin:
Hello,

Few changes for improve Windows compatibility:
- Windows with Python < 3.5 need `backports.socketpair` package. I added import for it  if `socket.socketpair` import fail.
- Changed `setup.py` for installing `backports.socketpair` package as requirement, but only for Windows and Python < 3.5.
- Changed path detection for correctly find `icons` and `locale` directories created by pip.

Also related change for the wiki : 

> ## For users of Windows and Mac OS X
> 
> *Note: these systems are in no way officially supported. Use ReText there at your own risk.*
> 
> ### For Windows
> simply install ReText with Pip using `pip install ReText` in cmd.exe, then download and unpack the [icon pack](https://sourceforge.net/projects/retext/files/Icons/) in `PythonFolder/share/icons` directory (`PythonFolder` is you Python install directory, where `python.exe` is).

> If you use Python < 3.5, the `backports.socketpair` package is additionally required to run ReText.

> _The old wiki has a [page documenting the procedure of installing ReText on Windows](https://sourceforge.net/p/retext/wiki/Windows Install of ReText/)_.

> ### For Mac OS X
> As there is no global icon theme on this systems, you need to download the [icon pack](https://sourceforge.net/projects/retext/files/Icons/) and unpack the icons from there into `ReText/icons` directory.
> 
> After that, running ReText locally (without installing) will work. If you want to install it, change the `icon_path = icons/` line in `ReText/__init__.py` to a different value (do not forget the trailing slash). If you want to use ReText locally, do not forget to install the needed dependencies using `pip`.
> 

And an extra suggestion: create a "ReTextIcons" package on Pypi containing icons for Windows/Mac and adding them to the `/share/icons` folder. Then adding it in `setup.py` as conditional suggestion like for `backports.socketpair`. I can eventually help for this.
With this, Windows users will stop to need extra manual interventions for installing ReText.

### DIFF
--- a/ReText/__init__.py
+++ b/ReText/__init__.py
@@ -22,6 +22,7 @@ from os.path import dirname, exists, join
 
 from PyQt5.QtCore import QByteArray, QLocale, QSettings, QStandardPaths
 from PyQt5.QtGui import QFont
+import platform
 
 app_version = "6.0.0"
 
@@ -33,6 +34,12 @@ if not str(settings.fileName()).endswith('.conf'):
 		'ReText project', 'ReText')
 
 datadirs = QStandardPaths.standardLocations(QStandardPaths.GenericDataLocation)
+
+if platform.system() == 'Windows':
+    # Windows compatibility: Add "PythonXXX\share\" path
+    import sys
+    datadirs.append(join(dirname(sys.executable), 'share'))
+
 datadirs = [join(d, 'retext') for d in datadirs]
 
 if '__file__' in locals():
--- a/ReText/converterprocess.py
+++ b/ReText/converterprocess.py
@@ -4,7 +4,11 @@ import markups
 import multiprocessing as mp
 import pickle
 import signal
-import socket
+try:
+    from socket import socketpair
+except ImportError:
+    # Windows compatibility : socket.socketpair backport for Python < 3.5
+    from backports.socketpair import socketpair
 import struct
 import traceback
 import weakref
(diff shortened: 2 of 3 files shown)


mitya57:
Thanks for your pull request. What's the issue with using Python 3.5? It's the latest stable version, released last year, and PyQt provides the [official Windows installers](https://riverbankcomputing.com/software/pyqt/download5) and [wheels](https://pypi.io/project/PyQt5/#files) only for Python 3.5 (not for 3.4). Why should we care about Python 3.4 then?

JGoutin:
Up to date Python is not always available in some companies.
Personally, I am stuck to 3.4.3 at job, and ask for a simple update is very long and costly...

Griffon26:
I sympathize with the issue of not having the latest version of things at a company.
For use at our company I created the https://github.com/retext-project/ReTextPortable project.
It includes python and other dependencies with the application and has a work-around to also use backports.socketpair, but obviously it's quite a big package as a result.

starofrainnight:
The same problem too. I have to work with windows 10 and python 3.4.3 and stuck to them. This PR is helpful for me, vote 1.

mitya57:
I don't want this compat code in master, but I can add it to `retext-6.0` branch so that it's included in 6.0.1 and subsequent 6.0.x releases, if there will be any.

mitya57:
I have pushed the (slightly updated) `converterprocess.py` and `__init__.py__` parts to both master and 6.0 branches. Thanks, and sorry for the delay.

I don't think the `setup.py` part is important enough, as the missing dependency is just one command away. (And Windows is not officially supported, especially with old Python versions).

PR #210, opened 2016-05-18
Fix dialog localization

JGoutin:
Hello,

Fix some dialogs buttons not localized, like "discard", "cancel", ...

Tested in french.


mitya57:
Thanks!

It appears that we don't actually need anything apart from qtbase, so I just renamed the existing path rather than adding a new one.

PR #211, opened 2016-05-20
Auto detect encoding

starofrainnight:
I have lots *.rst file in UTF-8 encoding, but while I open it, it display as jumbly characters until I “Set encoding" to "UTF-8" and reload it again. After go around the source codes, I found  retext will load the file with defaultCodec which be initialized to system encoding, but my system encoding is "GBK".

So if I load a file which content encoding different from defaultCodec, it will be displayed as jumbly characters until I set the encoding correct manually.

That's not the end: even I set the encoding correctly, all things seems display correctly, but after I save that file, it will be converted to defaultCodec (which is GBK). So now my UTF-8 encoding file be converted to a GBK encoding file.

I modified those routines about encoding parts:

1. Added encoding auto detect behavior before file loading and pass the detected encoding to QTextStream
2. Saved detected encoding to document object 
3. Save the file by detected encoding which buffered in the document object 

Now the file encoding load as is and save as it, and seems that we need not set the encoding manually.

This only tested on python3.4.3 on Win10 x64 with Chinese language.

### DIFF
--- a/ReText/tab.py
+++ b/ReText/tab.py
@@ -22,6 +22,8 @@ from markups.common import MODULE_HOME_PAGE
 from ReText import app_version, globalSettings, converterprocess
 from ReText.editor import ReTextEdit
 from ReText.highlighter import ReTextHighlighter
+import os
+import chardet
 
 try:
 	from ReText.fakevimeditor import ReTextFakeVimHandler
@@ -132,7 +134,7 @@ class ReTextTab(QSplitter):
 		'''
 		Set the default markup class to use in case a markup that
 		matches the filename cannot be found. This function calls
-		updateActiveMarkupClass so it can decide if the active 
+		updateActiveMarkupClass so it can decide if the active
 		markup class also has to change.
 		'''
 		self.defaultMarkupClass = markupClass
@@ -286,6 +288,22 @@ class ReTextTab(QSplitter):
 		self.editBox.setVisible(self.previewState < PreviewNormal)
 		self.previewBox.setVisible(self.previewState > PreviewDisabled)
 
+	def detectFileEncoding(self, fileName):
+		'''
+		Detect content encoding of specific file by first 512 bytes.
+
+		It will return the global default encoding if it can't determine
+		which the encoding is.
+		'''
+		someBytes = min(512, os.path.getsize(fileName))
+		raw = open(fileName, 'rb').read(someBytes)
+		result = chardet.detect(raw)
+
+		if result["confidence"] < 0.65:
+			return globalSettings.defaultCodec
+
+		return result['encoding']
+
 	def readTextFromFile(self, fileName=None, encoding=None):
 		previousFileName = self._fileName
 		if fileName:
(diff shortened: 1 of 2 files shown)


mitya57:
Thanks for your pull request.

Unfortunately it makes one of the tests fail (see Travis log), because apparently calling `setProperty` on a text document marks it as modified. Can you please fix it?

Bonus points for:

- Squashing the commits so that there are not so many of them;
- Removing the trailing whitespace (i.e. as highlighted by `git diff`).

starofrainnight:
Thanks for figure out which lead the CI failed, I will fix it soon.

Thanks your advices, I will make that comments more simple and clear and change my editor's settings for remove the trailing white spaces automatically.

starofrainnight:
@mitya57 

The problem was introduced by my first commit at f011a6a4fbcd779beab77a09ce5acf96d8c6268f.

That's not related to "setProperty()", setProperty() seems won't trigger the modified status (I have confirmed).

That's related to the old behavior which handle the modified status in ReTextTab.readTextFromFile().

Let's assume a scene and try to trigger this CI test problem:

1. Open ReText 
2. Open the existing_file.md in "tests/testdata"

The original ReTextTab.readTextFromFile

`
def readTextFromFile(self, fileName=None, encoding=None):
    previousFileName = self._fileName
    if fileName:
        self._fileName = fileName
    openfile = QFile(self._fileName)
    openfile.open(QFile.ReadOnly)
    stream = QTextStream(openfile)
    encoding = encoding or globalSettings.defaultCodec
    if encoding:
        stream.setCodec(encoding)
    text = stream.readAll()
    openfile.close()

    modified = bool(encoding) and (self.editBox.toPlainText() != text)
    self.editBox.setPlainText(text)
    self.editBox.document().setModified(modified)

    if previousFileName != self._fileName:
        self.updateActiveMarkupClass()
        self.fileNameChanged.emit()
`

I now change only one line from 

`encoding = encoding or globalSettings.defaultCodec`

to 

`encoding = 'ascii'`

and do the test to see what's happening (In my first commit, this change is the same what I have done), it failed the same just like the CI done.

Why this happening?

As we know globalSettings.defaultCodec is empty string by default, and will only pass valid encoding to readTextFromFile()  while user set encoding manually.

So the follow codes in readTextFromFile() will mostly return False (encoding mostly equal to None, so bool(encoding) = False):

`modified = bool(encoding) and (self.editBox.toPlainText() != text)`

Now the situation changed, because we now doing automatic encoding detect always, so the encoding mostly have something except it don't have enough confidence to determine the encoding then drop back to globalSettings.defaultCodec.

Then, bool(encoding) will mostly equal to True and in this assumed scene, it's equal to True.

The next condition "(self.editBox.toPlainText() != text)" will also be True, because original editBox have nothing in it but there have text in existing_file.md.

So finally, modified be changed to True every time we new open a file contained any text and the actionSave also be enabled this time.

Now how we fix it ?

Just remove the modified status confitions codes and set the modified status to False directly (See my 

I checked all code statements which invoked readTextFromFile(), and think this change won't harm those places, am I missing something?

I have committed this little changes, please check it again.

mitya57:
Right, I was wrong at guessing the reason. Thanks for finding the issue and fixing it!

I will look at this a bit later — maybe I'll redo some things in a bit different way.

starofrainnight:
@mitya57 

All commits are squashed.

mitya57:
@starofrainnight Thanks, though that was not needed because as said earlier I changed some things.

Summary of my changes:

- Made auto-detection optional (though enabled by default).
- Made `chardet` dependency optional too.
- Fixed resource leak when opening the file for read.
- Replaced min(512, length) with just 512 (there is no difference).
- Other minor code simplifications.

Thanks for your work again!

starofrainnight:
@mitya57 

Thanks your explanation! 
I'm not an expert of python, all experience are comes from C/C++.
I will learn from your changes and use more pythonic way, hope I won't waste your time on my next PR :)

Issue #135, opened 2015-02-03
Automatically reopen previous documents


mitya57:
Hello,

I have an idea for a future version of ReText. It would be fine to have an option to automatically reopen files which were open the last time. This does not seem to be available in the current version.
Would it be possible?

Thank you in advance and keep up the good work, it's fantastic!

Reported by: banasura

Original Ticket: [retext/tickets/135](https://sourceforge.net/p/retext/tickets/135)

rockiger:
The elementaryos hig have a good write up, why and how recreating state from last use makes sense:
https://elementary.io/en/docs/human-interface-guidelines#normal-launch
Hope this is useful

mitya57:
Pushed with minor changes and simplifications. Thanks!

PR #214, opened 2016-05-24
Auto reopen previous documents

starofrainnight:
This PR just want to add feature which described in #135 .

It added an option "openLastFilesOnStartup" to control that behavior, it will disabled by default.

### DIFF
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -20,7 +20,7 @@ import markups
 import sys
 from subprocess import Popen
 from ReText import icon_path, app_version, globalSettings, readListFromSettings, \
- writeListToSettings, writeToSettings, datadirs
+ readFromSettings, writeListToSettings, writeToSettings, datadirs
 from ReText.tab import ReTextTab, ReTextWebPreview, PreviewNormal, PreviewLive
 from ReText.dialogs import HtmlDialog, LocaleDialog
 from ReText.config import ConfigDialog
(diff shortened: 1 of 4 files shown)


ReText/window.py
```diff
@@ -353,6 +353,18 @@ def __init__(self, parent=None):
...
 
+		# Restore last opened files
+		if globalSettings.openLastFilesOnStartup:
+			files = readListFromSettings("lastFileList")
+			for file in files:
+				self.openFileWrapper(file)
+
+			# Trigger the tab of last opened file
+			lastTabIndex = readFromSettings(
+				"lastTabIndex", int, default=self.tabWidget.currentIndex())
```
mitya57: The `currentIndex` is (almost) always zero, so you can add `lastTabIndex` to the list of settings and use `globalSettings.lastTabIndex` instead.

ReText/window.py
```diff
@@ -1078,6 +1090,13 @@ def closeEvent(self, closeevent):
 				return closeevent.ignore()
 		if globalSettings.saveWindowGeometry and not self.isMaximized():
 			globalSettings.windowGeometry = self.saveGeometry()
+		if globalSettings.openLastFilesOnStartup:
+			files = []
+			for i in range(0, self.tabWidget.count()):
+				tab = self.tabWidget.widget(i)
+				files.append(tab.fileName)
+			writeListToSettings("lastFileList", files)
+			writeToSettings("lastTabIndex", self.tabWidget.indexOf(self.currentTab), int)
```
mitya57: … and `globalSettings.lastTabIndex = self.tabWidget.currentIndex()` here.

What you wrote is wrong because the third argument of `writeToSettings` is the default value, not the type.

Also, note that in my suggestion I replaced `indexOf()` with `currentIndex()` which is simplier.

starofrainnight:
@mitya57 

Already done your suggestions :

1. Moved lastTabIndex to  the list of settings and use globalSettings.lastTabIndex instead. So the writeToSettings() is unneeded now .
1. Replaced indexOf() with currentIndex()

mitya57:
Pushed with minor changes and simplifications. Thanks!

Issue #148, opened 2015-08-19
Open documents into new tab rather than new window


mitya57:
When I already have a ReText window open, I'd like to be able to double-click a file in the file explorer (Gnome Files) and have it open into a new tab on the existing window. Currently a new ReText window is opened instead. Thanks!

I'm using retext-5.0.2-2 on Fedora Linux 22.

Reported by: porjo38

Original Ticket: [retext/tickets/148](https://sourceforge.net/p/retext/tickets/148)

Mte90:
I developed a bit with PyQt and I can see that exist QtDbus also in python so I think that is possible to expose a signal.
In that way is possible to do a tiny script (or retext itself) that call dbus from retext and accept a file and will open a new tab.

mitya57:
A pull request adding that would be very welcome.

Mte90:
I am studying the code to see if I can do it.

I think that at https://github.com/retext-project/retext/blob/master/ReText/__main__.py#L92 I can see to emit the dbus signal when retext is started again.
Instead to the dbus listener I can use `window.openFileWrapper(fileName)` to open the file.
I can do it probably in the main file but I will let you know for updates in the next days.

Mte90:
I did it https://github.com/retext-project/retext/pull/476 :-D

I wrote a book with ReText (mentioned in tools used) and this is to say thank you to the project :-) https://github.com/retext-project/retext/pull/476

PR #215, opened 2016-05-27
Open documents in new tab 

starofrainnight:
I think the title should be named as "Open documents in new tab when another ReText starts from console".

This PR just want to help complete #148 .

Tested under : 

1. python3.4 on win10 x64 
2. python3.5.1 on ubuntu16.04 x64


retext.py
```diff
@@ -54,6 +57,39 @@ def main():
 	app.setOrganizationDomain('mitya57.me')
 	if hasattr(app, 'setDesktopFileName'): # available since Qt 5.7
 		app.setDesktopFileName('me.mitya57.ReText.desktop')
+
+	singleApp = SingleApplication("D1278E7822F011E6802800F1F38F93EF")
```
mitya57: What does this number mean? :)

mitya57:
@Griffon26 You have more experience with sockets and multiprocessing, can I please ask you to review this?

retext.py
starofrainnight: The number just an unique ID those all ReText shared, so that the new starting ReText will use to check if there already an ReText existed. It could be any text, but you must ensure it's unique value in system using by QSharedMemory (The key value of QSahredMemory).

The value in source code is generated by python code on my system:

`
import uuid
uuid.uuid1().hex.upper()
`

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+		self._name = name
+		self._mode = self.Server
+		self._sharedMemory = QSharedMemory(self._name)
+		self._systemSemaphore = QSystemSemaphore(self._name + "-2", 1)
+		self._server = None
+		self._client = None
+		self._localSockets = {}
+
+		# Fix for *nix:
+		# http://stackoverflow.com/questions/5006547/qt-best-practice-for-a-single-instance-app-protection
```
Griffon26: Keep the links, but add an explanation of the fix here, because understanding the code should not depend on a web site.

retext.py
```diff
@@ -54,6 +57,39 @@ def main():
...
+	singleApp.start()
+	if singleApp.mode == singleApp.Client:
+		# When we in client mode, we just send the arguments to existed
```
mitya57: When we **are** in client mode

retext.py
```diff
@@ -54,6 +57,39 @@ def main():
...
+		# appliaction and exit
```
mitya57: to the exist**ing** appli**ca**tion instance

retext.py
```diff
@@ -54,6 +57,39 @@ def main():
...
+		message = json.dumps(list(map(canonicalize, sys.argv[1:])))
+		singleApp.sendMessage(message.encode('utf-8'))
+		
+		# Exit the application immediately after send message
```
mitya57: after send**ing** the message

retext.py
```diff
@@ -72,6 +108,8 @@ def main():
 		app.setStyleSheet(QTextStream(sheetfile).readAll())
 		sheetfile.close()
 	window = ReTextWindow()
+	# Mark this window so that we could find it while we received message.
```
mitya57: wh**en** we received the message

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
+# This file is part of ReText
+# Copyright: 2012-2015 Dmitry Shachnev
```
mitya57: Please put your copyright here :)

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import functools
+import struct
+from PyQt5.QtCore import pyqtSignal
```
mitya57: Please group the imports, i.e.:

    from PyQt5.QtCore import pyqtSignal, QObject, QSharedMemory, QSystemSemaphore

retext.py
mitya57: How about just using "ReText" as the string?

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+			localSocket = self._server.nextPendingConnection()
+			self._localSockets[localSocket] = b""
+			localSocket.readyRead.connect(functools.partial(
+				self._onLocalSocketReadyRead, localSocket))
+
+	def start(self):
+		# Ensure we run only one application
+		isAnotherRunning = False
+		if not self._sharedMemory.isAttached():
+			# If share memory is attachable, that means there have another
```
mitya57: share**d** memory
there **exists** another running application

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+		if not isAnotherRunning:
+			# If we already attached by previous attach test, we will create
+			# the share memory.
+			self._systemSemaphore.acquire()
+			try:
+				isAnotherRunning = not self._sharedMemory.create(1, QSharedMemory.ReadWrite)
+			finally:
+				self._systemSemaphore.release()
+
+			# If there something happened during create procedure lead us can't
```
mitya57: Sorry, I can't parse this sentence…

If <s>there</s> something (what? maybe an example?) happened during the creation procedure and we can't create one (what?), we should detach that share**d** memory.

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+			# create one, we should detach that share memory.
+			if isAnotherRunning:
+				self._systemSemaphore.acquire()
+				try:
+					if self._sharedMemory.isAttached():
+						self._sharedMemory.detach()
+				finally:
+					self._systemSemaphore.release();
+
+		# Fine, now we could take different action on different situation.
```
mitya57: This comment is not informative at all — if you want to comment here, describe *which* action we can take.

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+		if not isAnotherRunning:
```
mitya57: I prefer `if…else` blocks rather than `if not…else`.

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+			# memory after original instance exitted while there still have
+			# ReText as client.
+
+			self._mode = self.Client
+			self._client = QLocalSocket(self)
+			self._client.connectToServer(self._name)
+
+	def sendMessage(self, message):
+		# Only accept bytes message
+		assert(type(message) == bytes)
```
mitya57: Was this assert just a debug, or really needed?
Also, `isinstance(message, bytes)` looks a bit more nice.

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+
+	# Signals
+
+	receivedMessage = pyqtSignal(bytes)
+
+	def __init__(self, name, parent=None):
+		QObject.__init__(self, parent)
+		self._name = name
+		self._mode = self.Server
+		self._sharedMemory = QSharedMemory(self._name)
```
Griffon26: Each user on the system should be able to run this application, but currently the code forces one instance per system.
It would be easy enough to include the username here, but that would still make it very easy for a malicious user to perform a denial of service attack on others. Can you think of a way around this?

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+				finally:
+					self._systemSemaphore.release();
+
+		# Fine, now we could take different action on different situation.
+		if not isAnotherRunning:
+			self._mode = self.Server
+			self._server = QLocalServer(self)
+			self._server.newConnection.connect(self._onServerNewConnection)
+
+			if not self._server.listen(self._name):
```
Griffon26: I think we need QLocalServer::UserAccessOption here to avoid one user forcing another user's ReText to open a file.
Just like with the semaphore/shared memory we will also need a way to avoid denial of service. Is there a portable way to use a user-specific namespace, such as a file in the user's homedir?

ReText/singleapplication.py
```diff
@@ -0,0 +1,167 @@
...
+import functools
+import struct
+from PyQt5.QtCore import pyqtSignal
+from PyQt5.QtCore import QObject
+from PyQt5.QtCore import QSharedMemory
+from PyQt5.QtCore import QSystemSemaphore
+from PyQt5.QtNetwork import QLocalServer
+from PyQt5.QtNetwork import QLocalSocket
+
+class SingleApplication(QObject):
```
Griffon26: Given the complexity of solving this problem in a portable way and the risk of introducing problems that opens the application up to interference from other users, I'm wondering if there isn't some existing proven code that we can just reuse. Have you looked at alternatives and ruled them out?

retext.py
starofrainnight: OK, It will be changed to "ReText", I just concentrated upon unique value at the beginning.
Later it will be changed to something like "ReText-username" for multi-user that @Griffon26 referred.

starofrainnight:
WOW, that's so many questions and so many changes I needs to do ...
This PR is so rough to just make the function works.
Actually I'm not **very** clear about the details how it works on each platform, I trying to figure them out.
That need some time before I reply all your questions (I have to take more research and clear my mind), hope I have time to done all by near two weeks. 

First thing, I will try to fix my grammar errors, thanks your patient @mitya57. Seems that's the easier things I could do now.

ReText/singleapplication.py
starofrainnight: OK, I will add an explanation here until I sorted them out later.

ReText/singleapplication.py
starofrainnight: I can't answer these questions yet, because the whole single application framework is converted from some Qt/C++ principle code snippets that I still not fully understand. I just combine them together, make it seems works and try my best to comment them.

Why I comments here with "something" just I really don't know what will make this procedure failed.

They related to different system parts, and at the time those code snippets don't have much details about why they do that.

starofrainnight: Maybe we don't need to comment here to tell what's going on, because we must do the clear job if the function failed to create, right?

I means, no matter what it happening, maybe system error or else, it must be clear after failed, so, we clear it.

ReText/singleapplication.py
starofrainnight: It's really needed.
I will change to `isinstance()` and raise TypeError if it failed.

starofrainnight: It's really needed.
I will change `isinstance()`  and raise TypeError() if failed

mitya57:
There is a [single application](http://code.qt.io/cgit/qt-solutions/qt-solutions.git/tree/qtsingleapplication) solution from Qt developers themselves. It relies on lock files so the application will have one instance per user, not per system. Also it uses QLocalSocket for messages exchange.

It was written before [QLockFile](http://doc.qt.io/qt-5/qlockfile.html) became part of Qt. Now that it's available, we can use it to simplify the task of porting that code to Python.

starofrainnight:
@mitya57 

I had go through that solution from Qt developers before this PR, it's seems too complicated to introduce a dozen classes into this project, so I left it for another. At that time, I don't know QLockFile has became part of Qt yet. Until you just refer to it.

@mitya57 @Griffon26 
I found a python single application implementation from tendo module, seems that if we introduce this module will solve our single application problem. It also relies on lock files. 

https://github.com/pycontribs/tendo/blob/master/tendo/singleton.py

There have two ways to rewrote this PR:

1. Rewrote by QLockFile
2. Introduce tendo module dependence and rewrote by it's SingleInstance class (I think this is more easy to implement)

Which one do you prefer?

Griffon26:
Be sure to check the license of the tendo code. To me it's not clear if it falls under the python license and if so, which version. Not all python license versions are compatible with the GPL ReText is under.

mitya57:
There are discussions in Qt upstream about including QSingleApplication in qtbase: http://lists.qt-project.org/pipermail/development/2016-June/026228.html

Let's see where it goes.

starofrainnight:
Yep, the license of the tendo code is not compatible with the GPL, so this way is down.
I trying reimplement by QLockFile.

mitya57:
I think the security issue mentioned by @Griffon26 in https://github.com/retext-project/retext/pull/215#discussion_r66705329 is still there: the server is available for all users, and one can perform a DoS attack on it which can lead to data loss.

Let's maybe wait for QSingleApplication to become reality instead.

Don't take me wrong, but I started ReText as a simple and lightweight editor, and I didn't really expect it to get lots of code with potential security issues written not by me…

starofrainnight:
@mitya57 

You are right, the security issue still there, I still looking into it. But I have too busy to take that investigation these days. 

OK, lets wait for QSingleApplication to become reality instead.

I understand your serious attitude to the code, actually I like the way you are.

PR #217, opened 2016-06-01
Improve auto encoding detect quality

starofrainnight:
The auto encoding detect feature detected 'ascii' encoding from ReText's "README.md", but that file's correct encoding is "utf-8".

The first importance character could be use to decide which encoding the file is appeared at 584 Bytes, that's exceeded our first detect area.

So now I return 'utf-8' encoding if the file be detected as 'ascii', because 'utf-8' encoding is compatible with 'ascii'.

At the same time, enlarged the detect area to 2048 Bytes for get more precise results.


mitya57:
Thanks!

Issue #220, opened 2016-06-02
Cursor is very hard to see


moorepants:
I find the cursor for text editing the input file is very difficult to see. It would be nice if it was a different color or a bit thicker. Even when I and moving it around with the arrow keys, I am often unable to detect the movement. It blends to well into the rst text.

Griffon26:
Can you provide a screenshot that shows what the cursor looks like on your system?

I would prefer to fix this problem without introducing a configuration item.
Configuration items should be reserved for personal preferences where it's too hard to find middle ground. After all perfection is attained not when there is nothing more to add, but when there is nothing more to take away.

What do you think, @mitya57?

moorepants:
![selection_014](https://cloud.githubusercontent.com/assets/276007/15782356/c6a88546-295e-11e6-9b3b-f569fcd64e2d.jpg)

Here is a screenshot. Ubuntu 16.04.

starofrainnight:
@Griffon26 

I agree what @moorepants said, that's boring to find the cursor, it too thin to see where it is.
I think that's because my screen is in High-Resolution (1600x900 with 14inch LCD), and the QTextEditor default cursor width is only 1 pixel ! 
That's hard to see it until enlarge the cursor width to at least 3 pixels.
Same situation happen on 19inch LCD with 1920x1080 resolution (My PC in company)

moorepants:
My resolution is 1920x1080 on a latop screen, so that's part of the problem. I don't seem to have this problem with cursors in other programs though. For example, typing this text box in Firefox is fine.

starofrainnight:
The cursor is the same thin as firefox's text box here, but it's more easy to find the cursor on firefox's text box.

I think the difference between them is that there have less characters in firefox's text box, but we will input a great lots text in ReText and the whole screen text will confuse our eyes.

starofrainnight:
@moorepants 

I just notice that, in firefox there have two different type cursors, one for address bar / search bar, it's as thin as retext, another is the web page editors, that one just like you said, it have 2 pixels width (a bit thicker).

starofrainnight:
@Griffon26

Maybe you could just set the default cursor width to 2 ~ 3 pixels so that without introducing a configuration item.
At least it seems notable while the cursor width be set to 3 pixels in my situation.

mitya57:
[...]

PR #221, opened 2016-06-03
Added option to control the cursor width

starofrainnight:
So you could control how thick the cursor for more easy find it in the editor.
Related to #220


starofrainnight:
Somehow, I agree with @Griffon26. 
No more option and this PR have other better solution.

PR #278, opened 2017-04-20
pt-PT translation fully finished

Ricardo-Simoes:
Portuguese European pt-PT localization fully finished


mitya57:
Muito obrigado!

Ricardo-Simoes:
De nada ;)

PR #283, opened 2017-05-16
Close current tab with Ctrl+W

xgouchet:
This adds the ability to quickly close an open tab from a keyboard shortcut without having to use the mouse


mitya57:
I would rather implement it as a global shortcut (i.e. add an action to the menu). This way it will work even in the preview mode when the editor is not focused.

xgouchet:
Good idea, done !

mitya57:
Merged with some changes. Thanks again!

PR #298, opened 2017-07-17
Handle quote blocks when handling return

xgouchet:
Handle quote block the same way lists are handle : a first return key will automatically fill the line with a starting `>`, and a second one will replace the line with a blank one.


mitya57:
Thank you!

PR #304, opened 2017-08-04
read/write desktop file with encoding utf-8

korli:
the content of the file me.mitya57.ReText.desktop is utf-8 encoded, so use that instead of the platform default.


mitya57:
Thank you!

PR #313, opened 2017-09-08
Implement softwrapping at user specified margin line

oguzhanogreden:
As I started using ReText more often, I noticed that I missed the softwrap option of Atom.

This pull request implements a simple softwrap feature and creates the necessary preferences entry. Moreover, it can take into account whether the width of ReTextEdit(QTextEdit) is smaller than user specified margin.

### DIFF
--- a/ReText/config.py
+++ b/ReText/config.py
@@ -84,7 +84,7 @@ class ConfigDialog(QDialog):
 			(self.tr('Tab key inserts spaces'), 'tabInsertsSpaces'),
 			(self.tr('Tabulation width'), 'tabWidth'),
 			(self.tr('Draw vertical line at column'), 'rightMargin'),
-            (self.tr('Enable soft wrap'), 'rightMarginWrap'),
+			(self.tr('Enable soft wrap'), 'rightMarginWrap'),
 			(self.tr('Show document stats'), 'documentStatsEnabled'),
 			(self.tr('Interface'), None),
 			(self.tr('Icon theme name'), 'iconTheme'),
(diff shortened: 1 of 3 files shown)


ReText/editor.py
```diff
@@ -106,12 +106,20 @@ def __init__(self, parent):
...
+		self.setWrapModeAndWidth()
 		self.document().blockCountChanged.connect(self.updateLineNumberAreaWidth)
 		self.cursorPositionChanged.connect(self.highlightCurrentLine)
 		self.document().contentsChange.connect(self.contentsChange)
 		if globalSettings.useFakeVim:
 			self.installFakeVimHandler()
 
+	def setWrapModeAndWidth(self):
+		if globalSettings.rightMarginWrap and (self.rect().topRight().x() > self.marginx):
+			self.setLineWrapMode(2)
```
mitya57: Please do not use magic numbers 1 or 2. Use `QTextEdit.WidgetWidth` and `QTextEdit.FixedPixelWidth` instead.

ReText/config.py
```diff
@@ -122,6 +124,8 @@ def initWidgets(self):
 			if isinstance(value, bool):
 				self.configurators[name] = QCheckBox(self)
 				self.configurators[name].setChecked(value)
+				if name == 'rightMarginWrap' and getattr(globalSettings, 'rightMargin') == 0:
```
mitya57: Same here, just use `globalSettings.rightMargin`.

mitya57:
Thanks you! Merged.

PR #319, opened 2017-09-12
Few typos fixed for Croatian translation

krunose:


mitya57:
Thank you!

There was no need to close pull requests, as you could use `git push --force` to update existing ones.

Issue #114, opened 2014-03-10
[spell check] Ignore/Add words to dictionary feature is missing


mitya57:
I would be really helpfull to have an Ignore/Add words to dictionary feature when using spell check.

Reported by: and3k

Original Ticket: [retext/tickets/114](https://sourceforge.net/p/retext/tickets/114)

mitya57:
Fixed in #320.

PR #320, opened 2017-09-13
Add unknown word to dictionary

### LINKED ISSUE #260: Feature Requset: Spellchecker ignore or add to dictionary
As I'm writing technical documents, getting spelling underlines on pppd, udev, idProduct, idVendor, and the like are somewhat distracting. But I want to know when I type "adn" instead of "and", so I don't want to entirely disable spell checking.

A lot of apps that have spell checking provide a way to ignore incorrect words (perhaps only for as long as the app is open), or add the words to a custom dictionary. Could you add similar options to the context menu in retext?

Thanks for retext, it's just plain awesome (fakeVIM, live preview, highlighting, rst, md, etc = just awesome).

xgouchet:
Add an entry in the context menu on a mispelled word to add the word to
the user dictionary.

Fixes #260, #114

### DIFF
--- a/ReText/editor.py
+++ b/ReText/editor.py
@@ -181,12 +181,29 @@ class ReTextEdit(QTextEdit):
 		actions = [self.parent.act(sug, trig=self.fixWord(sug)) for sug in suggestions]
 		menu = self.createStandardContextMenu()
 		menu.insertSeparator(menu.actions()[0])
+		menu.insertAction(menu.actions()[0], self.parent.act(self.tr('Add to dictionary'),
+			trig=self.learnWord(word)))
+		menu.insertSeparator(menu.actions()[0])
 		for action in actions[::-1]:
 			menu.insertAction(menu.actions()[0], action)
 		menu.exec(event.globalPos())
 
 	def fixWord(self, correctword):
 		return lambda: self.insertPlainText(correctword)
+
+	def learnWord(self, newword):
+		return lambda: self.addNewWord(newword)
+
+	def addNewWord(self, newword):
+		cursor = self.textCursor()
+		block = cursor.block()
+		cursor.clearSelection()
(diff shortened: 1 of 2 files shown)


ReText/editor.py
```diff
@@ -181,13 +181,28 @@ def contextMenuEvent(self, event):
...
+
+	def addNewWord(self, newword):
+		cursor = self.textCursor()
+		cursor.clearSelection()
+		self.setTextCursor(cursor)
+		dictionary = self.tab.highlighter.dictionary
+		if (dictionary is None) or not newword:
+			return
+		dictionary.add(newword)
+		self.tab.highlighter.rehighlight()
```
mitya57: Rehighlighting the whole document is slow. Maybe you can use [rehighlightBlock](http://doc.qt.io/qt-5/qsyntaxhighlighter.html#rehighlightBlock)?

ReText/editor.py
```diff
@@ -181,13 +181,28 @@ def contextMenuEvent(self, event):
 		actions = [self.parent.act(sug, trig=self.fixWord(sug)) for sug in suggestions]
 		menu = self.createStandardContextMenu()
 		menu.insertSeparator(menu.actions()[0])
+		menu.insertAction(menu.actions()[0], self.parent.act(self.tr('Add to dictionary'), trig=self.learnWord(word)))
```
mitya57: This line is too long, please break it into two smaller ones (PEP 8 says lines should be < 80 characters, but I am trying to at least keep them below 100).

xgouchet: Done !

ReText/editor.py
xgouchet: Done

mitya57:
:+1:

PR #326, opened 2017-09-21
Handle return in ordered lists context

xgouchet:
This allows to continue an ordered list by pressing return, the same way
unordered lists and quote blocks are handled.

I changed the detection to use a regex that might ease other supports

### DIFF
--- a/ReText/editor.py
+++ b/ReText/editor.py
@@ -93,6 +93,8 @@ def documentIndentLess(document, cursor, globalSettings=globalSettings):
 class ReTextEdit(QTextEdit):
 	resized = pyqtSignal(QRect)
 	scrollLimitReached = pyqtSignal(QWheelEvent)
+	returnBlockPattern = re.compile("^[\\s]*([*>-]|\\d+\\.) ")
+	orderedListPattern = re.compile("^([\\s]*)(\\d+)\\. $")
 
 	def __init__(self, parent):
 		QTextEdit.__init__(self)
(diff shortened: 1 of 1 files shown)


ReText/editor.py
```diff
@@ -242,15 +242,25 @@ def handleReturn(self, cursor):
...
-		pos = 0
-		while pos < length and (text[pos] in (' ', '\t')
-		  or text[pos:pos+2] in ('* ', '- ', '> ')):
-			pos += 1
-		if pos == length:
-			cursor.removeSelectedText()
-		# Reset the cursor
-		cursor = self.textCursor()
-		cursor.insertText(('\n' + text[:pos]) if pos < length else '\n')
+		match = re.search("^[\\s]*([*>-]|\\d+\\.) ", text)
```
mitya57: Please pre-compile the regexes for performance using re.compile.

ReText/editor.py
```diff
@@ -242,15 +242,25 @@ def handleReturn(self, cursor):
...
+		if match is not None:
+			matchedText = match.group(0)
+			if len(matchedText) == length:
+				cursor.removeSelectedText()
+				matchedText = ''
+			else :
```
mitya57: Please remove spaces after `else`.

xgouchet: Done :+1:

ReText/editor.py
xgouchet: Done :+1:

ReText/editor.py
mitya57: Sorry but here it is not done…

ReText/editor.py
```diff
@@ -105,6 +105,8 @@ def __init__(self, parent):
 		self.infoArea = LineInfoArea(self)
 		self.statistics = (0, 0, 0)
 		self.statsArea = TextInfoArea(self)
+		self.returnBlockPattern = re.compile("^[\\s]*([*>-]|\\d+\\.) ")
+		self.orderedListPattern = re.compile("^([\\s]*)(\\d+)\\. $")
```
mitya57: I would move it to class level (below the `scrollLimitReached = pyqtSignal(QWheelEvent)` line) for even more performance :)

ReText/editor.py
xgouchet: Oups Sorry 'bout that

ReText/editor.py
xgouchet: Done :)

mitya57:
Merged. Thanks a lot!

PR #330, opened 2017-10-02
Use the WaveUnderline style to mark spellcheck errors

### LINKED ISSUE #329: When a large font size is used, misspelled words aren't underlined
I am using ReText 7.0.1 from the Debian package.

If I set the editor font size to 16, misspelled words are underlined:

large

If I increase it to 18 or larger, they are not underlined anymore:

small

This is DejaVu Sans, but the issue seems to occur regardless of typeface.

xgouchet:
Fixes #329

It seems like this is a Qt bug, making the SpellCheckUnderline disappear
when the font is bigger than 16 pt. Any other underline style works
correctly.

There is an existing bug report on this :
https://bugreports.qt.io/browse/QTBUG-50499

### DIFF
--- a/ReText/highlighter.py
+++ b/ReText/highlighter.py
@@ -155,7 +155,7 @@ class ReTextHighlighter(QSyntaxHighlighter):
 		if self.dictionary:
 			charFormat = QTextCharFormat()
 			charFormat.setUnderlineColor(Qt.red)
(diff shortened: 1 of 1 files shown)


mitya57:
Thank you! I have an idea on how to fix this in Qt upstream, will probably submit a fix to them tomorrow.

mitya57:
Upd: submitted as https://codereview.qt-project.org/207461.

PR #334, opened 2017-10-16
Increase the bottom margin for text info

### LINKED ISSUE #333: Text statistics placement overlays last line in document
If document statistics are enabled they are displayed at the bottom of the text editor view.

This means the last line of text in the current viewport is obstructed by the statistics.

stats

I think it would be best to make the statistics not part of the text editor pane somehow or maybe make them have some margin to the text editor view.

Slightly related to #291

xgouchet:
This makes sure that the text info (current line, text statistics) do
not override the document's text.

Fixes #333

### DIFF
--- a/ReText/editor.py
+++ b/ReText/editor.py
@@ -289,7 +289,7 @@ class ReTextEdit(QTextEdit):
 		return 5 + self.fontMetrics().width('9') * digits
 
 	def updateLineNumberAreaWidth(self, blockcount=0):
-		self.setViewportMargins(self.lineNumberAreaWidth(), 0, 0, 0)
+		self.setViewportMargins(self.lineNumberAreaWidth(), 0, 0, self.fontMetrics().height())
 
 	def resizeEvent(self, event):
 		QTextEdit.resizeEvent(self, event)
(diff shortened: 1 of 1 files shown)


mitya57:
Sorry but I’m not a fan of this. My initial idea was having the info area that does not take any vertical space. It was semi-transparent and the text to the left of it was visible.

The statistics area did not play well with that design, but your change makes it even further from it.

xgouchet:
I understand your point of view. Do you see any other way to place the text statistics ? I thought maybe it could be displayed as a popup on demand (instead of having it displayed and computed all the time). What do you think?

mitya57:
Yes, that would be better. Or maybe move the statistics area to the right (there is usually less text on the right size), and additionally add some hack for the last line in the file hack to get #333 fixed?

xgouchet:
@mitya57 When you say move it to the right, do you imagine stacking the cursor info and stats info, or have them both on the right, one after the other ?

mitya57:
Maybe one above the other, maybe to the left from another. No strong opinion in this.

xgouchet:
Did the one above the other on the right side in #338

Issue #205, opened 2016-05-02
Change paper size for PDF export


stephenjjohnson:
The PDF export seems to default to A4, with no obvious (to me) way to change it.

I can print to PDF and change the paper to e.g. letter, and set up margins. Is there a UI or text file that stores default values for the more direct "Export" option?

mitya57:
There is no option to set the default paper size at the moment.

I can implement it if needed, though probably after 6.0 release.

stephenjjohnson:
It would be a nice feature. Either that, or persistent selections for paper size and margins in the print dialog. They have yet to embrace iso paper sizes here in the US. :-)

mitya57:
This was implemented in #335 and will be available in the next release.

You can use `paperSize` setting in the [configuration file](https://github.com/retext-project/retext/blob/master/configuration.md), e.g. `paperSize=Letter`.

PR #335, opened 2017-10-17
ENH: Add paperSize global setting 

mray271:
Adds paperSize global setting which sets both the default paper size for both the quick PDF export and the more interactive print dialog.

Reference issue #205

### DIFF
--- a/ReText/__init__.py
+++ b/ReText/__init__.py
@@ -98,7 +98,7 @@ configOptions = {
 	'useWebEngine': False,
 	'useWebKit': False,
 	'windowGeometry': QByteArray(),
-	'paperSize': 'A4',
+	'paperSize': '',
 }
 
 def readFromSettings(key, keytype, settings=settings, default=None):
--- a/ReText/__init__.py
+++ b/ReText/__init__.py
@@ -98,6 +98,7 @@ configOptions = {
 	'useWebEngine': False,
 	'useWebKit': False,
 	'windowGeometry': QByteArray(),
+	'paperSize': 'A4',
 }
 
 def readFromSettings(key, keytype, settings=settings, default=None):
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -950,23 +950,22 @@ class ReTextWindow(QMainWindow):
 				printer.setPaperSize(pageSize)
 			else:
 				QMessageBox.warning(self, '',
-					self.tr('Unrecognized PaperSize setting "%s".' %
-					globalSettings.paperSize))
+					self.tr('Unrecognized PaperSize setting "%s".') %
+					globalSettings.paperSize)
 		return printer
 
 	def getPageSizeByName(self, pageSizeName):
-		""" Returns a validated PageSize instance
-		corresponding to the given name. Returns
-		None if the name is not a valid PageSize.
+		""" Returns a validated PageSize instance corresponding to the given 
+        name. Returns None if the name is not a valid PageSize.
 		"""
-
+		
 		pageSize = None
-		match = None
-		if pageSizeName:
-			match = list(filter(lambda y:y.lower()==pageSizeName.lower(),
-						self.availablePageSizes()))
-		if match:
-			pageSize = getattr(QPagedPaintDevice, match[0])
+        
+		lowerCaseNames = {pageSize.lower(): pageSize for pageSize in 
+						  self.availablePageSizes()}
+		if pageSizeName.lower() in lowerCaseNames:
+			pageSize = getattr(QPagedPaintDevice, lowerCaseNames[pageSizeName])
+
 		return pageSize
 
 	def availablePageSizes(self):
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -944,29 +944,30 @@ class ReTextWindow(QMainWindow):
 		printer = QPrinter(QPrinter.HighResolution)
 		printer.setDocName(title)
 		printer.setCreator('ReText %s' % app_version)
-		page_size = self.getPageSizeByName(globalSettings.paperSize)
-		if page_size is not None:
-		    printer.setPaperSize(page_size)
-		else:
-		    QMessageBox.warning(self, '',
-				'Unrecognized PaperSize setting "{0}".'.format(
-				globalSettings.paperSize))
+		if globalSettings.paperSize:
+			pageSize = self.getPageSizeByName(globalSettings.paperSize)
+			if pageSize is not None:
+				printer.setPaperSize(pageSize)
+			else:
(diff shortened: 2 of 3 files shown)


ReText/window.py
```diff
@@ -943,8 +944,39 @@ def standardPrinter(self, title):
...
+		if page_size_name:
+			match = list(filter(lambda y:y.lower()==page_size_name.lower(),
+						self.availablePageSizes()))
+		if match:
+			page_size = getattr(QPagedPaintDevice, match[0])
+		return page_size
+
+	def availablePageSizes(self):
+		""" List available page sizes.
+		"""
```
mitya57: Maybe use [QPrinter::availablePageSizes()](http://doc.qt.io/qt-5/qprinterinfo.html#supportedPageSizes) instead of this function? You will need to pass printer info to `getPageSizeByName`, but that should not be a problem.

ReText/window.py
```diff
@@ -943,8 +944,39 @@ def standardPrinter(self, title):
...
+		None if the name is not a valid PageSize.
+		"""
+
+		page_size = None
+		match = None
+		if page_size_name:
+			match = list(filter(lambda y:y.lower()==page_size_name.lower(),
+						self.availablePageSizes()))
+		if match:
+			page_size = getattr(QPagedPaintDevice, match[0])
```
mitya57: I would use a dict here. Something like:
```python
sizes = {size.name().lower(): size for size in printerInfo.supportedPageSizes()}
requestedSize = sizes.get(pageSizeName.lower())
```

ReText/__init__.py
```diff
@@ -98,6 +98,7 @@ def getBundledIcon(iconName):
 	'useWebEngine': False,
 	'useWebKit': False,
 	'windowGeometry': QByteArray(),
+	'paperSize': 'A4',
```
mitya57: Why default explicitly to `A4`? Maybe leave the default empty and use Qt’s defaults?

mray271: Done

ReText/window.py
```diff
@@ -943,8 +944,39 @@ def standardPrinter(self, title):
 		printer = QPrinter(QPrinter.HighResolution)
 		printer.setDocName(title)
 		printer.setCreator('ReText %s' % app_version)
+		page_size = self.getPageSizeByName(globalSettings.paperSize)
+		if page_size is not None:
+		    printer.setPaperSize(page_size)
+		else:
+		    QMessageBox.warning(self, '',
+				'Unrecognized PaperSize setting "{0}".'.format(
```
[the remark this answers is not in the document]
mray271: Done

ReText/window.py
mray271: Hmm.  supportedPageSizes  (possibly same as availablePageSizes ?) consistently returns empty array in my set-up.  Perhaps this feature is not quite working on my system or there are some version differences.  I'm using python-qt5-5.7-7.fc25.x86_64 on Fedora  Linux 4.13.5-100.fc25.x86_64.  Leaving as is for now as this seems compatible/working. Feel free to modify/

ReText/window.py
mray271: Hmm.  supportedPageSizes (possibly same as availablePageSizes ?) consistently returns empty array in my set-up.  Perhaps this feature is not quite working on my system or there are some version differences. I'm using python-qt5-5.7-7.fc25.x86_64 on Fedora  Linux 4.13.5-100.fc25.x86_64. Leaving as is for now as this seems compatible/working. Feel free to modify/

ReText/window.py
```diff
@@ -943,8 +944,40 @@ def standardPrinter(self, title):
 		printer = QPrinter(QPrinter.HighResolution)
 		printer.setDocName(title)
 		printer.setCreator('ReText %s' % app_version)
+		if globalSettings.paperSize:
+			pageSize = self.getPageSizeByName(globalSettings.paperSize)
+			if pageSize is not None:
+				printer.setPaperSize(pageSize)
+			else:
+				QMessageBox.warning(self, '',
+					self.tr('Unrecognized PaperSize setting "%s".' %
```
mitya57: The closing `)` is in wrong place, this won’t work. The translators need to translate the string with literal `%s`, not with the substituted value.

It should be `self.tr('Unrecognized PaperSize setting "%s".') % globalSettings.paperSize`.

ReText/window.py
```diff
@@ -943,8 +944,40 @@ def standardPrinter(self, title):
...
+		"""
+
+		pageSize = None
+		match = None
+		if pageSizeName:
+			match = list(filter(lambda y:y.lower()==pageSizeName.lower(),
+						self.availablePageSizes()))
+		if match:
+			pageSize = getattr(QPagedPaintDevice, match[0])
+		return pageSize
```
mitya57: Can you please still use a dictionary? Then this function will be just 3 lines instead of current 8.

```python
lowerCaseNames = {pageSize.lower(): pageSize for pageSize in self.availablePageSizes()}
if pageSizeName.lower() in lowerCaseNames:
    return getattr(QPagedPaintDevice, lowerCaseNames[pageSizeName])
```

(And it will still return None if there is no match.)

ReText/window.py
```diff
@@ -943,8 +944,40 @@ def standardPrinter(self, title):
...
+				QMessageBox.warning(self, '',
+					self.tr('Unrecognized PaperSize setting "%s".' %
+					globalSettings.paperSize))
 		return printer
 
+	def getPageSizeByName(self, pageSizeName):
+		""" Returns a validated PageSize instance
+		corresponding to the given name. Returns
+		None if the name is not a valid PageSize.
+		"""
```
mitya57: Any reason why the docstring lines are so short?

ReText/window.py
mray271: Done.

ReText/window.py
mray271: Fixed those so they wrap at longer length.

ReText/window.py
mray271: Done.

mray271:
Yeah, I agree it's super weird that an empty array is being returned. I tried exactly this QPrinterInfo(printer).supportedPageSizes() both inside standardPrinter with the printer there as input and a couple of other places and in an ipython session, but strangely it's still returning [] (an empty array).  I'm equally surprised. I can try to dig a bit further, but I'm guessing something strange about python-qt5 (PyQt5) or some other Fedora-specific issue.

mitya57:
Congratulations with your first PR merged! There were minor problems with whitespace, but I fixed them myself in a follow-up commit.

PR #337, opened 2017-10-17
Add shortcut to move a line up and down

xgouchet:
Pressing Alt+Arrow Up / Alt+Arrow Down will move the current line
up/down. This is a nice feature to move lines without having to do a
Select-Cut-Move-Paste manipulation.

### DIFF
--- a/ReText/editor.py
+++ b/ReText/editor.py
@@ -173,35 +173,42 @@ class ReTextEdit(QTextEdit):
 		self.lineNumberArea.update()
 
 	def contextMenuEvent(self, event):
+		# Create base menu
+		menu = self.createStandardContextMenu()
 		text = self.toPlainText()
-		dictionary = self.tab.highlighter.dictionary
-		if (dictionary is None) or not text:
-			return QTextEdit.contextMenuEvent(self, event)
+		if not text:
+			menu.exec(event.globalPos())
+			return
+
+		# Check word under the cursor
 		oldcursor = self.textCursor()
 		cursor = self.cursorForPosition(event.pos())
-		pos = cursor.positionInBlock()
+		self.setTextCursor(cursor)
+		pos = cursor.position()
 		if pos == len(text):
 			pos -= 1
 		curchar = text[pos]
 		isalpha = curchar.isalpha()
-		cursor.select(QTextCursor.WordUnderCursor)
-		if not isalpha or (oldcursor.hasSelection() and
-		oldcursor.selectedText() != cursor.selectedText()):
-			return QTextEdit.contextMenuEvent(self, event)
-		self.setTextCursor(cursor)
-		word = cursor.selectedText()
-		if not word or dictionary.check(word):
-			self.setTextCursor(oldcursor)
-			return QTextEdit.contextMenuEvent(self, event)
-		suggestions = dictionary.suggest(word)
-		actions = [self.parent.act(sug, trig=self.fixWord(sug)) for sug in suggestions]
-		menu = self.createStandardContextMenu()
-		menu.insertSeparator(menu.actions()[0])
-		menu.insertAction(menu.actions()[0], self.parent.act(self.tr('Add to dictionary'),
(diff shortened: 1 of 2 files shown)


mitya57:
Maybe implement it as QAction instead? This way you will be able to show it in Edit menu and right-click menu for increased visibility.

xgouchet:
@mitya57 Done, I also added the actions in the tool bar

ReText/window.py
```diff
@@ -336,6 +343,9 @@ def __init__(self, parent=None):
 		self.editBar.addAction(self.actionCopy)
 		self.editBar.addAction(self.actionPaste)
 		self.editBar.addSeparator()
+		self.editBar.addAction(self.actionMoveUp)
+		self.editBar.addAction(self.actionMoveDown)
+		self.editBar.addSeparator()
```
mitya57: Is it really such a common action that it needs to be on the toolbar?

ReText/editor.py
```diff
@@ -246,6 +249,10 @@ def keyPressEvent(self, event):
 					# Insert Markdown-style line break
 					cursor.insertText('  ')
 				self.handleReturn(cursor)
+		elif key == Qt.Key_Down and event.modifiers() & Qt.AltModifier:
+			self.moveLineDown()
+		elif key == Qt.Key_Up and event.modifiers() & Qt.AltModifier:
+			self.moveLineUp()
```
mitya57: These lines are no longer needed, the shortcut is handled by QAction.

ReText/window.py
xgouchet: Not really, I can remove it if you prefer

ReText/editor.py
xgouchet: done

ReText/editor.py
```diff
@@ -197,6 +197,9 @@ def contextMenuEvent(self, event):
 		actions = [self.parent.act(sug, trig=self.fixWord(sug)) for sug in suggestions]
 		menu = self.createStandardContextMenu()
 		menu.insertSeparator(menu.actions()[0])
+		menu.insertAction(menu.actions()[0], self.parent.actionMoveDown)
+		menu.insertAction(menu.actions()[0], self.parent.actionMoveUp)
+		menu.insertSeparator(menu.actions()[0])
```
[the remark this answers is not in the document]
xgouchet: I used `insert*` instead of `add*` to match the other lines in this method. Should we use `add*` for all of them ? 
Also I can't move those lines up as the menu is not created yet. Or we should not use the default menu creation at all  in this case.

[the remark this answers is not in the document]
mitya57: `insert*(menu.actions()[0], ...)` adds an action top top, `add*(...)` adds to bottom. In my opinion spell checker actions are popular and should be on top, these ones are less popular and can be on bottom.

What do you mean by “not use the default menu creation at all”?

I think the solution is like this:
- Always create the menu on top of this function, always append the move up/down actions;
- Replace `return QTextEdit.contextMenuEvent(self, event)` with `menu.exec(event.globalPos())`.

mitya57:
By the way I have invited you to @retext-project, but please still make pull requests for non-trivial changes.

xgouchet:
Wow thanks man. And don't worry, I think I'll still make pull request anyway to get a second pair of eyes on my code !

ReText/editor.py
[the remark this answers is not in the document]
xgouchet: Done. I kinda rewrote the method to add the action even if we don't need the spellcheck, while keeping the spellcheck options on top, and the move at the bottom

xgouchet:
Indeed, this should be fixed now

xgouchet:
Done

holzkohlengrill:
Don't know if that was intended or not but you cannot move multiple lines (only one).

I am commenting here since I don't know if this would be a new feature (if yes I will add this as a new issue) or if this is just a bug.

PR #338, opened 2017-10-18
Move the statistics aerea to the right side of the editor

### LINKED ISSUE #333: Text statistics placement overlays last line in document

xgouchet:
Fixes #333


mitya57:
Thank you!

Issue #236, opened 2016-09-09
Whishlist: ASCIIMath for math (optional)


mlopezgva:
Hi!

Since MathJax is already used to allow mathematical expressions, I would like to use **ASCIIMath** (http://asciimath.org/) which is way much simpler (and more "markdown-like") than LaTeX.

Is that possible through some setting tweaking or it does require some addicional work?

Thanks.

mitya57:
It is currently not possible to use AsciiMath. Leaving the bug open because I may implement it in some future version.

mlopezgva:
Hi again...

Now I saw reText is written in python... :-)

ASCIIMath at its homepage is written in javascript, however, there are some implementation written in python. I hope one or some of them helps...

https://github.com/favalex/python-asciimathml
https://pypi.python.org/pypi/asciitomathml/.89
https://github.com/ShadowKyogre/python-asciimathml
http://www.freewisdom.org/projects/python-markdown/  (related to the former)

PR #339, opened 2017-10-23
Asciimath poc

mray271:
Adds a proof of concept for AsciiMath for issue #236 .

Since forward slash in markdown indicates inline code, but in AsciiMath indicates math markup, users would just need to escape their forward slashes (i.e. using \\` )

image

```
Code in LaTeX
    \begin{equation}
    \int_{a}^{b}x^{2}dx = \frac{1}{3}x^{3} + C
    \end{equation}


AsciiMath Example 1

\`[[a,b],[c,d]]\`

AsciiMath Example 2

\`sum_(i=1)^n i^3=((n(n+1))/2)^2\`
```

### DIFF
--- a/ReText/tab.py
+++ b/ReText/tab.py
@@ -203,9 +203,17 @@ class ReTextTab(QSplitter):
 			headers += ('<link rel="stylesheet" type="text/css" href="%s">\n'
 			% cssFileName)
 		headers += ('<meta name="generator" content="ReText %s">\n' % app_version)
-		return converted.get_whole_html(
+		html = converted.get_whole_html(
 			custom_headers=headers, include_stylesheet=includeStyleSheet,
 			fallback_title=baseName, webenv=webenv)
+		old_mathjax = '<script type="text/javascript" src="file:///usr/share/javascript/mathjax/MathJax.js"></script>'
+		new_mathjax = """
+			    	  <script type="text/javascript" async src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=AM_HTMLorMML">
+			    	  </script>\n
+					  """
+		html = html.replace(old_mathjax,new_mathjax)
+		
+		return html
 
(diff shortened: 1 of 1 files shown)


mitya57:
I can try to implement it properly in my [python-markdown-math extension](https://github.com/mitya57/python-markdown-math). All we need is output `<script type="math/asciimath">` tags, MathJax will take care of the rest.

Using ``\` `` as delimiter looks like a bad idea to me, since that means a literal backtick in Markdown. Maybe you have some suggestions for a better delimiter?

mitya57:
Or maybe I should keep the current delimiters and add an option to switch between TeX and ASCIIMath syntax?

mray271:
In LaTeX, inline math is delimited by $ as in $ x^{2} $.  So that might work.
Something similar for AsciiMath might be \`$[[a,b],[c,d]]\`$  Something similar apparently, was done here: 
https://discourse.gohugo.io/t/how-i-extended-markdown-for-hugo/5983
but maybe that has other issues.

Maybe \\\` is not "too bad" for AsciiMath since it is simple and close to what MathJax already uses and there do exist other ways for someone to encode literal backtick should they need:  https://meta.stackexchange.com/questions/55437/how-can-the-backtick-character-be-included-in-code#answer-238046

i)  \`\` \` \`\` (Opening and closing the code section with 2 or more backticks)
ii) \<code\>&#96\</code\>
iii) \<code\>\`\</code\>

mitya57:
I finally managed to get all parts of the stack working together. With the upcoming (hopefully today) PyMarkups and python-markdown-math releases, you will be able to use AsciiMath like this:

ReText screenshot

mitya57:
One more note: the markup does not allow having both LaTeX and AsciiMath math in the same file.

I don't see why one would need this, but if you really want that, you can always insert the math blocks manually using:
```html
<script type="math/tex">TeX markup here</script>
<script type="math/asciimath">AsciiMath markup here</script>
```

PR #375, opened 2018-06-11
Inherit `previewState`  on new tab while opening links

DaniloGomez:


mitya57:
Thanks for your pull request!

This change makes `previewState` inherited not only when clicking links, but also when simply opening files. Was this intended? If yes, I think this should be configurable with a setting.

DaniloGomez:
So, you are suggesting me to separate the proposed feature in two, lets say
1. `previewState` inheritance when clicking links
2. `previewState` inheritance when opening files

 and keep the feature 1 as a desired behavior and left the feature 2 configurable by a setting?
Or maybe both features should be configurable?

mitya57:
There is no need to separate this into two features. I was just trying to understand your initial goal.

- If you care only about clicking links, then please bring the code in accordance with the pull request description (don’t change behavior for opening files).
- If you care about opening files as well as clicking links, then it should be configurable.

mitya57:
I am closing this because you have not addressed my comment above. Please feel free to submit a new pull request if you are still on it.

PR #407, opened 2018-10-30
Add setting for number of recent documents

red-kite:


mitya57:
Thanks! Merged both PRs.

Issue #396, opened 2018-09-28
Images are not cached


dopeboy:
I want to start with the comment that this is a fantastic tool. Thank you @mitya57 and others for all your effort on this.

I was using retext on a low data cap connection. I had an image in my markdown file and noticed it gets reloaded in full when I make an update (of any kind) to my markdown file. Is there a way to cache images? If not, would you be open to a PR for it?

mitya57:
Currently the only way to cache image is to save it locally, which is what I recommend to do :)

I will be open to a pull request if the implementation is not too complex.

dopeboy:
Thanks for the quick reply, mitya57. I think my use case might be worth still solving if the final content is going to remotely hosted e.g. on github.

I'll perform a quick analysis and draft up an approach. If it sounds reasonable to you, I'll proceed with a PR.

dopeboy:
Here's a proposed approach. All of it refers to `tab.py`. Feedback is welcome.

* In `startPendingConversion()`, take `self.editBox.toPlainText()` and scan for images (both in markdown and HTML) that have a remote URL. For each such image, download the image and save it into `/tmp`. In this text that is being send to be rendered, replace the remote URLs with local paths.

* When the live preview is closed (can be closed manually or through closing the file), delete these files in `/tmp`. 

* In `triggerPreviewUpdate()`, take an optional parameter. Pass this optional parameter to `startPendingConversion()`. This optional parameter will be non-null when `triggerPreviewUpdate()` is invoked by a text change. This is useful because if the user updated their remotely hosted image, we want a way for them to force a deep refresh of the live preview so that their new image shows. Text changes shouldn't initiate that but, say, opening and closing the live preview area should.

dopeboy:
Bump @mitya57

mitya57:
This was fixed in 5c7b38916f0d2e60495175a9f77a1ff5df2a0e9b and confirmed in https://github.com/retext-project/retext/pull/409#issuecomment-458758594, so I am closing this issue.

PR #409, opened 2018-11-17
Cache remotely hosted images

dopeboy:
Spurred from #396, this PR caches referenced images that are remotely hosted. As of now, every keystroke initiates a remotely hosted image to be redownloaded. This can tax low data cap connections.

This first draft scans the text for HTML  tags that rely on remotely hosted images. Each time such an image is encountered, it gets saved to /tmp/ and added to a dictionary for tracking. When the tab or application is closed, these images are deleted. @mitya57, I'd love to get your feedback on this first take.

Next up is:
- Supporting markdown images
- Handling when there is no network connection or poor network connection (will crash right now)
- Writing some tests


ReText/tab.py
```diff
@@ -275,13 +283,19 @@ def startPendingConversion(self):
 			self.converterProcess.start_conversion(self.getActiveMarkupClass().name,
 			                                       self.fileName,
 			                                       requested_extensions,
-			                                       self.editBox.toPlainText(),
+			                                       self.cacheRemotelyHostedImages(self.editBox.toPlainText()),
```
mitya57: You are calling this on the source Markdown/reST file, not on the resulting HTML. The source file will usually not have `` tags.

ReText/tab.py
```diff
@@ -456,3 +470,47 @@ def openSourceFile(self, fileToOpen):
...
+
+			# If this <img/> tag is not remotely hosted, move on.
+			if img_src.startswith('http') is False:
+				continue
+
+			# If it is remotely hosted and we haven't seen it before,
+			# give it a unique filename and save it into /tmp/. Also, save it into
+			# a dict (seen) so that we can track it for later use and cleanup
+			if img_src not in self.seen:
+				local_file = f'/tmp/{str(uuid.uuid4())}_{img_src[img_src.rfind("/")+1:]}'
```
mitya57: `/tmp/` will not work on some platforms, such as Windows.

Please use a cross-platform solution such as [tempfile](https://docs.python.org/3/library/tempfile.html) module.

ReText/tab.py
```diff
@@ -456,3 +470,47 @@ def openSourceFile(self, fileToOpen):
...
+			text = f'{text[:start_position-offset]}{local_file}{text[end_position-offset:]}'
+			offset += len(img_src)-len(local_file)
+		return text
+
+	def cleanImageCache(self):
+		"""
+		This runs when a tab is closed OR when the application is closed.
+		"""
+		[os.remove(image['local_file']) for image in self.seen.values()]
+		self.seen = {}
```
mitya57: Please use a more specific variable name, such as `seen_images`.

ReText/tab.py
```diff
@@ -45,12 +51,14 @@
 	ReTextWebEnginePreview = None
 
 PreviewDisabled, PreviewLive, PreviewNormal = range(3)
+HTML_IMG_TAG_SRC_RE = re.compile(r'src[\s]*=[\"\']([^\"\'>]+)[\"\']\/?>')
```
mitya57: This will match not only `` tags, but also other tags with `src` attribute, such as `<script>`.

ReText/tab.py
```diff
@@ -456,3 +470,47 @@ def openSourceFile(self, fileToOpen):
...
+					'start_position': start_position,
+					'end_position': end_position}
+				urllib.request.urlretrieve(img_src, local_file)
+
+			# If we've seen this image before, pull it up where it is saved
+			else:
+				local_file = self.seen[img_src]['local_file']
+
+			# Replace the path of the tag with our locally saved file
+			text = f'{text[:start_position-offset]}{local_file}{text[end_position-offset:]}'
```
mitya57: ReText supports Python 3.2+. But formatted strings were introduced only in Python 3.6. Please use old-style string formatting for now.

ReText/tab.py
dopeboy: That's true but they can. So I think we would have to support them.

ReText/tab.py
dopeboy: Oh shoot, thanks! TIL'd and fixed.

ReText/tab.py
dopeboy: Done!

ReText/tab.py
dopeboy: You're right, my mistake. Fixed.

ReText/tab.py
dopeboy: This is what I get for not reading the README. Switched over to `format()`

dopeboy:
Thanks for your feedback @mitya57 - made some changes that should resolve things. Will proceed on the two latter bulletpoints if there aren't any other flags.

ReText/tab.py
mitya57: I mean you should parse the converted output, not the source output. See e.g. the `html` variable in `updatePreviewBox()` method.

Parsing Markdown with regular expressions is a bad idea as there are always corner cases that you will miss.

And there is also reStructuredText, Textile and other markups…

If you parse the converted HTML then there is only one language to parse.

mitya57:
Looks better now, I marked all my other comments as resolved.

ReText/tab.py
dopeboy: I see what you mean now. This should be resolved now.

dopeboy:
@mitya57 now includes handling no network and tests. I see the Travis build failing off the `xvfb-run` command but that does work locally for me; would you mind taking a look at that?

tests/test_window.py
```diff
@@ -397,6 +400,85 @@ def test_doesNotTweakSpecialCharacters(self):
...
+	    '![]({url})'.format(url=self.fakeRemoteUrl))
+        processEventsUntilIdle()
+        time.sleep(0.5)
+        processEventsUntilIdle()
+
+        # Check that the cache is non-empty
+        self.assertTrue(self.window.currentTab.seenImages)
+
+        # Check that we see a temporary file generated in the preview box text
+        text = self.window.currentTab.previewBox.page().mainFrame().toHtml()
```
mitya57: ReText has three preview backends, based on QTextBrowser, Qt WebKit and Qt WebEngine.

The default one is QTextBrowser, so it should be used in tests. I have fixed this now.

dopeboy:
Sorry for the late reply here as well, coming back from holiday.

I haven't tried your code yet; my original observation came from monitoring my network usage. I've attached a video showing more - does not that happen on your end?

https://drive.google.com/file/d/16XNQCUF56WSEDwW03RMtvYUoeg8IvV4c/view?usp=sharing

mitya57:
Do you use Qt WebKit or Qt WebEngine renderer? (Check what option is present/selected in the Edit menu.)

dopeboy:
You are right, WebEngine does cache images. WebKit does not. The reason why I didn't use WebEngine to start with is because it doesn't preview my images:

image

Given that the WebEngine is experimental and not the default renderer, I think this PR still has some merit. Happy to attach `if WebKit enabled` guards around it. Will defer to your guidance here.

mitya57:
> You are right, WebEngine does cache images. WebKit does not. The reason why I didn't use WebEngine to start with is because it doesn't preview my images

You have some misunderstanding here.

ReText has *three* preview backends, based on QTextBrowser (the default one), Qt WebKit and Qt WebEngine (Chromium).

Qt WebKit is deprecated upstream. If you are using an official Qt build (e.g. from the PyQt5 wheel), it will not have the Qt WebKit module. However some Linux distributions, such as Debian/Ubuntu, still ship it. If Qt WebKit is installed, ReText will have an option in the Edit menu to use it. If it is not installed, but Qt WebEngine is, ReText will have an option for WebEngine.

QTextBrowser does not display external images at all, it renders only local images (in fact I recommend to save images locally anyway). So cache is not applicable to this backend.

Qt WebEngine uses cache by default, as I illustrated in my example above.

Finally, Qt WebKit supports cache, but apparently does not use it by default. I have just implemented basic cache support and pushed it to `webkitpreview-cache` branch, see the linked commit. I this this solution is cleaner than your proposed mechanism: it has less lines of code, and it properly supports HTTP `Cache-Control` / `Expires` headers. Can you please test it?

I have not yet implemented clearing the cache when page is closed, but I can try to implement it too.

dopeboy:
Thanks for taking the time to explain that. I've just tested the new branch it is great. Agreed, this is a far more elegant solution.

Closing this one down.

mitya57:
Thanks for confirming! I will merge my branch a bit later (still I'm very limited in time).

PR #413, opened 2018-12-01
README.md - add link to https://facelessuser.github.io/pymdown-extensions/

nodiscc:


mitya57:
This may create a false illusion that extensions on that page are the only supported ones.

Maybe link to https://github.com/retext-project/retext/wiki/Markdown-extensions instead, and advertize pymdown-extensions on that wiki page?

nodiscc:
@mitya57 , nevermind, I just realized this page could be reached by following the links trough https://github.com/retext-project/retext/blob/master/README.md -> https://github.com/retext-project/retext/wiki -> https://github.com/retext-project/retext/wiki/Markdown-extensions -> https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions -> https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions#bundles -> https://github.com/facelessuser/pymdown-extensions

They are not obvious to find, but I think it works OK this way.

Thank you, closing then.

PR #417, opened 2018-12-09
Add OARS metadata

scx:
I would like to add [OARS](https://hughsie.github.io/oars) rating to the AppData file.

See also:
https://github.com/flathub/flathub/wiki/App-Requirements#appstream
https://hughsie.github.io/oars/generate.html


mitya57:
Thanks!

PR #429, opened 2018-12-21
Add option to show full path in titlebar

red-kite:
There might be a more lean solution for the construct
QFileInfo(self.currentTab.fileName).filePath()

Option takes effect on changing tabs

### DIFF
--- a/ReText/__init__.py
+++ b/ReText/__init__.py
@@ -92,6 +92,7 @@ configOptions = {
 	'styleSheet': '',
 	'syncScroll': True,
 	'tabBarAutoHide': False,
+	'titleBarFullPath': False,
 	'tabInsertsSpaces': True,
 	'tabWidth': 4,
 	'uiLanguage': QLocale.system().name(),
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -462,6 +462,9 @@ class ReTextWindow(QMainWindow):
 		if tab == self.currentTab:
 			if tab.fileName:
 				self.setWindowTitle("")
+				if globalSettings.titleBarFullPath:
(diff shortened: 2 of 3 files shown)


red-kite:
btw -- while doing this I noticed to see the full file-path also as tab-tooltip :-)

ReText/window.py
```diff
@@ -462,6 +462,9 @@ def tabFileNameChanged(self, tab):
 		if tab == self.currentTab:
 			if tab.fileName:
 				self.setWindowTitle("")
+				if globalSettings.titleBarFullPath:
+					fullpath = QFileInfo(self.currentTab.fileName).filePath()
```
mitya57: `self.currentTab.fileName` should be already a canonical file path, it is set here:

https://github.com/retext-project/retext/blob/748a1041a48f4b5c95f6b3f586a761cfa32721e9/ReText/window.py#L821

red-kite:
I don't urgently need this and therefore implemented with an option.

ReText/__init__.py
```diff
@@ -92,6 +92,7 @@ def getBundledIcon(iconName):
 	'styleSheet': '',
 	'syncScroll': True,
 	'tabBarAutoHide': False,
+	'titleBarFullPath': False,
```
mitya57: Can you please keep this file sorted? I.e. the new option should go after `tabWidth`.

Also please document this new option in `configuration.md`.

red-kite: I'll try to do better in future

red-kite:
'window title' is far better and unambiguous. Thank you.

PR #435, opened 2019-01-08
Add option to set default preview state

red-kite:
Simultaneously remove obsolete option for default live-preview

New tabs get additional triggerPreviewUpdate


mitya57:
Thanks for the pull request!

I have changed the setting to be string, and changed the labels to be better readable.

Also I have added a line to make sure empty document is not created in preview mode. It has a drawback that if you set mode to normal-preview, and then open a document in the existing (empty) tab, it will be open in edit mode (since the createTab() method is not called). But we can solve it later if you want.

red-kite:
> Also I have added a line to make sure empty document is not created in preview mode.

*Great thought!* (already had this in my mind for another commit)
Your implementation works for *New File* but existing, empty files open in preview.

In window.py:
``` python
from os import stat

		#if previewState == PreviewNormal and not fileName:
		if previewState == PreviewNormal and (not fileName or stat(fileName).st_size == 0):
```
does the trick for me.

I could follow your mentioned drawback by opening a file from the menu in a new tab. I have never done this bevor. Think this is a very special case we can live with for now.

Thank you very much!

mitya57:
I think opening existing, empty files is also a very special use case. And you can always disable preview for them manually.

red-kite:
As I'm already using "Ask to create missing files #436" extensively this happens quite often. But for the current issue this doesn't matter. Close accepted :-)

PR #436, opened 2019-01-09
Ask to create missing files

red-kite:
After some struggeling with a statusBar solution I utilized the already used QMessageBox to get user's permission to create a linked, but nonexistent file. The file is subsequently opend.

### DIFF
--- a/ReText/tab.py
+++ b/ReText/tab.py
@@ -486,7 +486,7 @@ class ReTextTab(QSplitter):
 		:rtype: bool
 		"""
 		buttonReply = QMessageBox.question(self, self.tr('Create missing file?'),
-			self.tr("The file '%s' does not exist.\n\nDo you want to create it?" % fileToCreate),
+			self.tr("The file '%s' does not exist.\n\nDo you want to create it?") % fileToCreate,
 			QMessageBox.Yes | QMessageBox.No, QMessageBox.No)
 		if buttonReply == QMessageBox.Yes:
 			if self.createFile(fileToCreate):
@@ -508,12 +508,12 @@ class ReTextTab(QSplitter):
 			# 'No such file or directory' reduces to 'No such directory'
 			# because we call createFile if not exists(fileToOpen)
 			QMessageBox.warning(self, 'No such directory',
-								self.tr("File could not be created: '%s'" % fileToCreate))
+								self.tr("File could not be created: '%s'") % fileToCreate)
 			return False
 		except IOError as errormsg:
 			# Does ReText support Python <3.3? If not change IOError to OSError.
 			QMessageBox.warning(self, 'Warning',
-								self.tr('File could not be created: %s' % errormsg))
+								self.tr('File could not be created: %s') % errormsg)
 			# Printing errormsg within general OSError avoids individual
 			# handling of PermissionError and possibly other.
 			return False
--- a/ReText/tab.py
+++ b/ReText/tab.py
@@ -32,7 +32,7 @@ except ImportError:
 
 from PyQt5.QtCore import pyqtSignal, Qt, QDir, QFile, QFileInfo, QPoint, QTextStream, QTimer, QUrl
 from PyQt5.QtGui import QTextCursor, QTextDocument
-from PyQt5.QtWidgets import QTextEdit, QSplitter
+from PyQt5.QtWidgets import QTextEdit, QSplitter, QMessageBox
 
 try:
 	from ReText.webkitpreview import ReTextWebKitPreview
@@ -447,6 +447,12 @@ class ReTextTab(QSplitter):
 		# TODO: broken objective? See TODO below.
 		It has to be located next to the current opened file.
 		Relative paths like [test](../test) or [test](folder/test) are also possible.
+
+		:param fileToOpen:	Link target
+							This can be a file of any extension or a path
+		:returns: 			Filename successfully handled to openFileWrapper
+							or 'noSuchFile' to stop further processing.
+		:rtype: 			str
 		"""
 		if self.fileName:
 			# Please give description of this condition.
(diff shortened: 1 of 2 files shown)


red-kite:
Works only for files handled by ReText. Creating also other formats like .txt and more to be opened with other software could be a future issue.

mitya57:
Sorry, I do not like this idea. If you want to create a new file, then why not just use the New button and then Save? This will avoid useless dialogs.

What other text editors do this?

red-kite:
I'm using ReText mainly as note-taking software with the superb ability to crosslink to all kind of data. Wether it's extern data like Websites or local documents like other textfiles, pdfs or anything – it can easily be opend from within Retext. *Thumbs up*, other markup editors I tried cannot do comparatively.

I'm proposing this feature because I do often create a link to a file that is to be filled with content later on. This technique also works very well with relative links. I suppose it can be thus useful for all kind of writing multipaged documentation.

mitya57:
Sorry for delay.

> I'm proposing this feature because I do often create a link to a file that is to be filled with content later on. This technique also works very well with relative links.

Ah, I understand now. Then please move the logic to `openSourceFile` method of `ReTextTab`. The `openFileWrapper` method is too generic — it is used for opening files via Open dialog, opening recent files on startup, etc. Then you won't need your special case for handling missing files on startup.

ReText/window.py
```diff
@@ -823,6 +826,17 @@ def openFile(self):
 	def openFileWrapper(self, fileName):
 		if not fileName:
 			return
+		if not QFile.exists(fileName):
+			if self.createFileQuestion(fileName):
+				try:
+					print('Create file' + ' ' + fileName , file=sys.stdout)
+					open(fileName, 'x').close()
+				except:
```
mitya57: Please do not use bare `except:`, always specify exceptions you want to catch.

See https://docs.quantifiedcode.com/python-anti-patterns/correctness/no_exception_type_specified.html

ReText/window.py
```diff
@@ -823,6 +826,17 @@ def openFile(self):
 	def openFileWrapper(self, fileName):
 		if not fileName:
 			return
+		if not QFile.exists(fileName):
+			if self.createFileQuestion(fileName):
+				try:
+					print('Create file' + ' ' + fileName , file=sys.stdout)
```
mitya57: If this is debug print, delete it. If you think it's useful then please:

- Remove `file=sys.stdout`, it is default anyway.
- Use a format function, like `'Create file %s' % fileName`, for readability.

ReText/window.py
```diff
@@ -823,6 +826,17 @@ def openFile(self):
...
+					open(fileName, 'x').close()
+				except:
+					QMessageBox.warning(self, 'Warning', 'File could not be created.')
+					return
+			else:
+				print('User decided not to create nonexisting file.', file=sys.stdout)
```
mitya57: Same as above.

ReText/window.py
```diff
@@ -847,6 +861,16 @@ def openFileWrapper(self, fileName):
 			self.currentTab.readTextFromFile(fileName)
 			self.moveToTopOfRecentFileList(self.currentTab.fileName)
 
+	def createFileQuestion(self, fileToCreate):
+		buttonReply = QMessageBox.question(self, self.tr('Create missing file?'),
+			self.tr('The file') + ' \'' + fileToCreate + '\' ' + self.tr('does not exist.')
```
mitya57: Pass the full string to translators. E.g.:
```python
self.tr("The file '%s' does not exist. Do you want to create it?") % fileToCreate
```

ReText/window.py
```diff
@@ -395,7 +395,10 @@ def __init__(self, parent=None):
 
 	def restoreLastOpenedFiles(self):
 		for file in readListFromSettings("lastFileList"):
-			self.openFileWrapper(file)
+			if QFile.exists(file):
+				self.openFileWrapper(file)
+			else:
+				print("Note: File '" + file + "' has been deleted!")
```
mitya57: This code will be no longer needed if you follow my advice.

red-kite:
Thanks a lot! I appreciate your suggestions for improvement. Currently into other tasks but will try to maintain your ideas…

ReText/tab.py
```diff
@@ -444,15 +444,76 @@ def openSourceFile(self, fileToOpen):
...
+
+		if get_markup_for_file_name(fileToOpen, return_class=True):
+			if not QFile.exists(fileToOpen):
+				# ReText should only support to create markup-files, no plain
+				# text files. Thereby it facilitates writing cross-linked
+				# documentation.
+				# Consequently create-file functionality can be nested here.
+				if not self.createFileQuestion(fileToOpen):
+					# File still not exists
+					return 'noSuchFile'
```
mitya57: What if the file name is `noSuchFile`? Maybe better to `raise FileNotFoundError`?

ReText/tab.py
```diff
@@ -444,15 +444,76 @@ def openSourceFile(self, fileToOpen):
...
+		except FileNotFoundError:
+			# Most expected error due possible misspelling in path and
+			# should thus be handled separately.
+			# 'No such file or directory' reduces to 'No such directory'
+			# because we call createFile if not exists(fileToOpen)
+			QMessageBox.warning(self, 'No such directory',
+								self.tr("File could not be created: '%s'" % fileToCreate))
+			return False
+		except IOError as errormsg:
+			# Does ReText support Python <3.3? If not change IOError to OSError.
```
mitya57: We support Python 3.2 and later currently (see README). You can change this in a separate pull request if you want.

ReText/tab.py
```diff
@@ -444,15 +444,76 @@ def openSourceFile(self, fileToOpen):
...
 			return fileToOpen
+
+	def createFileQuestion(self, fileToCreate):
+		"""Ask user if file should be created and do so if answer is yes.
+
+		:returns: True if new file has been created, False otherwise
+		:rtype: bool
+		"""
+		buttonReply = QMessageBox.question(self, self.tr('Create missing file?'),
+			self.tr("The file '%s' does not exist.\n\nDo you want to create it?" % fileToCreate),
```
mitya57: Here and in other places: you should pass **unsubstituted** string to `tr()`. E.g. not
```python
self.tr("The file '%s' does not exist.\n\nDo you want to create it?" % fileToCreate)
```
but rather
```python
self.tr("The file '%s' does not exist.\n\nDo you want to create it?") % fileToCreate
```

Otherwise translating just won't work.

ReText/window.py
```diff
@@ -847,6 +847,7 @@ def openFileWrapper(self, fileName):
 			self.currentTab.readTextFromFile(fileName)
 			self.moveToTopOfRecentFileList(self.currentTab.fileName)
 
+
```
mitya57: There should be just one blank line between methods of the same class, not two.

ReText/tab.py
red-kite: Tested raising an error but that started real trouble. My intention was to return 'something' to trigger the return-statement in (webkit/webengine)preview.py

My understanding is that the returnvalue is never processed and [return fileToOpen] could be replaced with [return 'filesuccessfullyopened'] without breaking something.

ReText/tab.py
```diff
@@ -444,15 +444,76 @@ def openSourceFile(self, fileToOpen):
...
+		:rtype: 			str
 		"""
 		if self.fileName:
+			# Please give description of this condition.
+			# (e.g. "if internal link was clicked from within valid open document in a tab"?)
 			currentExt = splitext(self.fileName)[1]
 			basename, ext = splitext(fileToOpen)
 			if ext in ('.html', '') and exists(basename + currentExt):
+				# TODO: It seems this should catch markdown links like
+				#  [test](test) but than ext not in ('.html', '')
```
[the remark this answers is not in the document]
red-kite: My comment on this was simply wrong, sorry for that.

mitya57:
Sorry for the delay in reviewing this. This week @xgouchet opened #459 which is inspired by your PR with some further refactoring, so I merged that instead.

Thanks to you both again for this contribution!

PR #459, opened 2019-05-20
Create missing file

xgouchet:
Offer the user to create missing file when following link.  
This is inspired by @red-kite PR #436

I split the openSourceFile method in two, to separate the "find"
responsibility from the "open" responsibility.


mitya57:
Merged with some minor changes. Thanks!

PR #476, opened 2019-10-06
Fix #148 - Open documents into new tab rather than new window

### LINKED ISSUE #148: Open documents into new tab rather than new window

Mte90:
In this way if there is already ReText instance, the second one send the file to the other and kill himself.

### DIFF
--- a/ReText/__main__.py
+++ b/ReText/__main__.py
@@ -32,6 +32,7 @@ from PyQt5.QtCore import QCommandLineOption, QCommandLineParser, QFile, \
  QFileInfo, QIODevice, QLibraryInfo, QTextStream, QTranslator, Qt
 from PyQt5.QtWidgets import QApplication
 from PyQt5.QtNetwork import QNetworkProxyFactory
+from PyQt5.QtDBus import QDBusConnection, QDBusInterface
 
 def canonicalize(option):
 	if option in ('--preview', '-'):
@@ -74,7 +75,10 @@ def main():
 	parser.addVersionOption()
 	previewOption = QCommandLineOption('preview',
 		QApplication.translate('main', 'Open the files in preview mode'))
+	newWindowOption = QCommandLineOption('new-window',
+		QApplication.translate('main', 'Create a new window even if there is an existing one'))
 	parser.addOption(previewOption)
+	parser.addOption(newWindowOption)
 	parser.addPositionalArgument('files',
 		QApplication.translate('main', 'List of files to open'),
 		'[files...]')
@@ -100,6 +104,20 @@ def main():
 		app.setStyleSheet(QTextStream(sheetfile).readAll())
 		sheetfile.close()
 	window = ReTextWindow()
+
+	connection = QDBusConnection.sessionBus()
+	if connection.isConnected() and not parser.isSet(newWindowOption):
+		connection.registerObject('/', window, QDBusConnection.ExportAllSlots)
+		serviceName = 'me.mitya57.ReText'
+		if not connection.registerService(serviceName) and filesToOpen:
+			print('Opening the file(s) in the existing window of ReText.')
+			iface = QDBusInterface(serviceName, '/', '', connection)
+			for fileName in filesToOpen:
+				iface.call('openFileWrapper', fileName)
+			qWidgetIface = QDBusInterface(serviceName, '/', 'org.qtproject.Qt.QWidget', connection)
+			qWidgetIface.call('raise')
+			sys.exit(0)
+
 	window.show()
 	# ReText can change directory when loading files, so we
 	# need to have a list of canonical names before loading
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -41,7 +41,7 @@ except ImportError:
 	enchant = None
 
 from PyQt5.QtCore import QDir, QFile, QFileInfo, QFileSystemWatcher, \
- QIODevice, QLocale, QMarginsF, QTextCodec, QTextStream, QTimer, QUrl, Qt
+ QIODevice, QLocale, QMarginsF, QTextCodec, QTextStream, QTimer, QUrl, Qt, pyqtSlot
 from PyQt5.QtGui import QColor, QDesktopServices, QIcon, \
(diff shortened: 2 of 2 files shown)


ReText/window.py
```diff
@@ -826,6 +826,7 @@ def openFile(self):
 		for fileName in fileNames[0]:
 			self.openFileWrapper(fileName)
 
+	@pyqtSlot(str)
```
Mte90: With this is possible to expose other methods of the UI on dbus, in case there are other needs

ReText/__main__.py
```diff
@@ -92,6 +93,18 @@ def main():
 	fileNames = list(map(canonicalize, sys.argv[1:]))
 	previewMode = False
 	readStdIn = False
+    
+	if not QDBusConnection.sessionBus().registerService('org.ReText'):
```
mitya57:
```suggestion
	if not QDBusConnection.sessionBus().registerService('me.mitya57.ReText'):
```

Let’s be consistent with the appdata and desktop files.

ReText/__main__.py
```diff
@@ -92,6 +93,18 @@ def main():
...
+		print('Send file to main instance')
+		iface = QDBusInterface('org.ReText', '/', '', QDBusConnection.sessionBus())
+		msg = iface.call('openFileWrapper', sys.argv[1] if len(sys.argv) > 1 else "")
+		QDBusReply(msg)
+		sys.exit(1)
+
+	connection = QDBusConnection.sessionBus()
+	connection.registerObject('/', window, QDBusConnection.ExportAllSlots)
+	connection.registerService('org.ReText')
```
mitya57: The connection is already registered on line 97, isn’t it? Can we leave only one of these lines?

And maybe also define `connection` earlier and don’t call `sessionBus()` twice.

ReText/__main__.py
```diff
@@ -92,6 +93,18 @@ def main():
...
+    
+	if not QDBusConnection.sessionBus().registerService('org.ReText'):
+		print('Send file to main instance')
+		iface = QDBusInterface('org.ReText', '/', '', QDBusConnection.sessionBus())
+		msg = iface.call('openFileWrapper', sys.argv[1] if len(sys.argv) > 1 else "")
+		QDBusReply(msg)
+		sys.exit(1)
+
+	connection = QDBusConnection.sessionBus()
+	connection.registerObject('/', window, QDBusConnection.ExportAllSlots)
```
mitya57: This also exports some slots on `org.qtproject.Qt.QWidget` interface that should not be exposed publicly. Like `setWindowModified` and `setWindowTitle`. Is there any way to avoid that?

ReText/__main__.py
```diff
@@ -92,6 +93,18 @@ def main():
 	fileNames = list(map(canonicalize, sys.argv[1:]))
 	previewMode = False
 	readStdIn = False
+    
+	if not QDBusConnection.sessionBus().registerService('org.ReText'):
+		print('Send file to main instance')
+		iface = QDBusInterface('org.ReText', '/', '', QDBusConnection.sessionBus())
+		msg = iface.call('openFileWrapper', sys.argv[1] if len(sys.argv) > 1 else "")
+		QDBusReply(msg)
+		sys.exit(1)
```
mitya57: I think it's not a failure, so better to use exit code 0 instead of 1.

Mte90:
Right now is missing:

* I don't know how to add a new parameter to ReText, do you have any suggestion where is the code? I see that manually you are checking the sys.argv, but for new parameters is not a scalable way
* To expose only the method of `ReTextWindow` that are not part of `QWindow` we need to create a new QDbusAdapter with the list of all the methods. Maybe we can create a wrapper class that include only the ones that we really want.

mitya57:
Thanks for your work! I am on a vacation now, will finish the implementation myself next week.

Mte90:
No problem @mitya57 :-)

mitya57:
Merged now, thank you!

ReText/__main__.py
```diff
@@ -74,7 +75,10 @@ def main():
 	parser.addVersionOption()
 	previewOption = QCommandLineOption('preview',
 		QApplication.translate('main', 'Open the files in preview mode'))
+	newWindowOption = QCommandLineOption('new-window',
```
Mte90: I didn't know of this class in Qt, my fault

ReText/__main__.py
```diff
@@ -100,6 +104,20 @@ def main():
...
+	connection = QDBusConnection.sessionBus()
+	if connection.isConnected() and not parser.isSet(newWindowOption):
+		connection.registerObject('/', window, QDBusConnection.ExportAllSlots)
+		serviceName = 'me.mitya57.ReText'
+		if not connection.registerService(serviceName) and filesToOpen:
+			print('Opening the file(s) in the existing window of ReText.')
+			iface = QDBusInterface(serviceName, '/', '', connection)
+			for fileName in filesToOpen:
+				iface.call('openFileWrapper', fileName)
+			qWidgetIface = QDBusInterface(serviceName, '/', 'org.qtproject.Qt.QWidget', connection)
```
Mte90: uh okay so you are exposing only a class and not all of them

ReText/__main__.py
```diff
@@ -100,6 +104,20 @@ def main():
 		app.setStyleSheet(QTextStream(sheetfile).readAll())
 		sheetfile.close()
 	window = ReTextWindow()
+
+	connection = QDBusConnection.sessionBus()
+	if connection.isConnected() and not parser.isSet(newWindowOption):
+		connection.registerObject('/', window, QDBusConnection.ExportAllSlots)
```
Mte90: so you are exposing all of them

mitya57: Yes, I am still exposing all of them, and taking an opportunity to use this, by calling the `raise()` method.

PR #480, opened 2019-10-28
Display * in tab title when document is modified and unsaved

### LINKED ISSUE #479: Show unsaved documents in tab title
When a document is modified, but the changes are not saved, the title of the window changes from `document.md` to `document.md*`. It would be nice when multiple documents are open to also add the star on tabs containing documents with unsaved modifications

xgouchet:
Fixes #479

### DIFF
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -455,7 +455,7 @@ class ReTextWindow(QMainWindow):
 
 	def updateTabTitle(self, ind, tab):
 		changed = tab.editBox.document().isModified()
-		if changed:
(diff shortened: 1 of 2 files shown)


mitya57:
Thank you!

I removed all `updateTabTitle()` calls after `tab.saveTextToFile()`. The latter emits `fileNameChanged` signal, so the handler of that signal will update the title anyway.

Also one of the calls was broken because `tab` was not defined (which made the tests fail).

xgouchet:
Yes thanks for that. I saw the test failed but didn't get a chance to update my pr myself. 👍

PR #494, opened 2020-02-17
Fix AppData file

scx:
XML declaration is allowed only at the start of the document.

Some XML tools (e.g. `xmlstarlet`) cannot process the file without this patch.

```
-:2.6: XML declaration allowed only at the start of the document
<?xml version="1.0" encoding="UTF-8"?>
     ^
-:1.1: Document is empty
```


mitya57:
Thanks!

PR #500, opened 2020-03-27
Insert an image directly from filesystem

DanielVenturini:
Hello, nice software.

I guess the retex would be great if it allows insert image directly by filesystem. The first commit allows to open a select box and get the image path, that is, selecting an image results in:
`Alt text`.

The second commit create a menu bar called **Insert** with the *Tables* and *Image* as sub-menus, where *Image* uses the function written at first commit. The *Tables* sub-menus (*Insert table* and *Table editing mode*) were in **Edit** menu, and it doesn't make much sense, I guess. Then, I removed those and inserted in **Insert** menu.

### DIFF
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -1254,11 +1254,10 @@ class ReTextWindow(QMainWindow):
 
 		cursor = self.currentTab.editBox.textCursor()
 
-		# if there're more than one file, add break lines '\n'
-		bl = '\n' if len(fileNames) > 1 else ''
-		for fileName in fileNames:
-			imageMarkup = self.currentTab.editBox.getImageMarkup(fileName)
-			cursor.insertText(imageMarkup + bl)
+		imagesMarkup = '\n'.join(
+			self.currentTab.editBox.getImageMarkup(fileName)
+			for fileName in fileNames)
+		cursor.insertText(imagesMarkup)
 
 		self.formattingBox.setCurrentIndex(0)
 		self.currentTab.editBox.setFocus(Qt.OtherFocusReason)
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -1248,15 +1248,17 @@ class ReTextWindow(QMainWindow):
 	def insertImage(self):
 		supportedExtensions = ['.png', '.jpg', '.jpeg', '.gif', '.bmp']
 		fileFilter = ' (' + str.join(' ', ['*' + ext for ext in supportedExtensions]) + ');;'
-		fileName = QFileDialog.getOpenFileName(self,
-			self.tr("Select one or several files to open"), QDir.currentPath(),
+		fileNames = QFileDialog.getOpenFileNames(self,
+			self.tr("Select one or several images to open"), QDir.currentPath(),
 			self.tr("Supported files") + fileFilter + self.tr("All files (*)"))
 
-		fileName = fileName[0]
-		imageText = '![{0}]({1})'.format(self.tr('Alt text'), self.tr(fileName))
-
 		cursor = self.currentTab.editBox.textCursor()
-		cursor.insertText(imageText)
+		# if there're more than one file, add break lines '\n'
+		bl = '\n' if len(fileNames[0]) > 1 else ''
+		for fileName in fileNames[0]:
+			imageText = '![{0}]({1}){2}'.format(self.tr('Alt text'), self.tr(fileName), bl)
+			cursor.insertText(imageText)
+
 		self.formattingBox.setCurrentIndex(0)
 		self.currentTab.editBox.setFocus(Qt.OtherFocusReason)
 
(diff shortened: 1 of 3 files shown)


ReText/window.py
```diff
@@ -324,13 +327,15 @@ def __init__(self, parent=None):
...
 		menuEdit.addAction(self.actionPreview)
-		menuEdit.addAction(self.actionInsertTable)
-		menuEdit.addAction(self.actionTableMode)
 		if ReTextFakeVimHandler:
 			menuEdit.addAction(self.actionFakeVimMode)
 		menuEdit.addSeparator()
 		menuEdit.addAction(self.actionFullScreen)
 		menuEdit.addAction(self.actionConfig)
+		menuInsert.addAction(self.actionInsertTable)
+		menuInsert.addAction(self.actionTableMode)
```
mitya57: I think that table mode has nothing to do with insertion, better leave it where it was.

And because of that, the Insert menu is becoming too small, so I would make it a submenu of Edit.

ReText/window.py
```diff
@@ -217,6 +217,8 @@ def __init__(self, parent=None):
 			lambda: self.searchBar.setVisible(False),
 			shct=QKeySequence.Cancel)
 		self.actionCloseSearch.setPriority(QAction.LowPriority)
+		self.actionInsertImage = self.act(self.tr('Image'), 'image',
```
mitya57: This will be easy to confuse with the existing “Paste image” option, which allows pasting from clipboard.

Maybe rename to something like “Image by file path”?

ReText/window.py
```diff
@@ -217,6 +217,8 @@ def __init__(self, parent=None):
...
+										  self.insertImage)
```
mitya57: Please use tabs for indentation but spaces for alignment, so

```suggestion
		                                  self.insertImage)
```

ReText/window.py
```diff
@@ -1239,6 +1247,21 @@ def viewHtml(self):
...
 
+	def insertImage(self):
+		supportedExtensions = ['.png', '.jpg', '.jpeg', '.gif', '.bmp']
+		fileFilter = ' (' + str.join(' ', ['*' + ext for ext in supportedExtensions]) + ');;'
+		fileName = QFileDialog.getOpenFileName(self,
+												 self.tr("Select one or several files to open"), QDir.currentPath(),
+												 self.tr("Supported files") + fileFilter + self.tr("All files (*)"))
+
+		fileName = fileName[0]
+		imageText = '![{0}]({1})'.format(self.tr('Alt text'), self.tr(fileName))
```
mitya57: Please also add syntax for reStructuredText and Textile. There is this block in ReText/editor.py:
https://github.com/retext-project/retext/blob/27a2745e3ae65bf894f359c5e05fe0559a65a99d/ReText/editor.py#L416-L422

Probably it should be split out to a separate function/method and reused here.

ReText/window.py
DanielVenturini: Ok, right

ReText/window.py
DanielVenturini: Such as to insert image will go to Edit menu, I will put as "Insert images by file path"

ReText/window.py
DanielVenturini: Sure!

ReText/window.py
mitya57: OK. Maybe in singular? “Insert *image* by file path”.

DanielVenturini: Sure. But now, I'm using the `getOpenFileNames` method to allows get more than one image. Still use *image* in singular?

mitya57: In this diff, you are using `getOpenFileName`, not `getOpenFileNames`…

DanielVenturini:
So, I revert my first commit, that is, I undo the **Insert** menu. Then, the `Insert images by file path` is in **Edit** menu and uses `getOpenFileNames` to allows the user add more than one image at same import. For the last, I splet into a function to get the correct syntax and turned it in a *static method* to access in both files.

ReText/window.py
```diff
@@ -1105,6 +1109,9 @@ def insertFormatting(self, formatting):
 			dialog.show()
 			self.formattingBox.setCurrentIndex(0)
 			return
+		elif formatting == 'image':
+			self.insertImage()
+			return
```
DanielVenturini: And what about it. This makes the `Formatting -> image` doesn't insert the `Alt text`, but opens the select image box instead. Do you thing is better remove it to insert the `Alt text`, or keeps it to open the select image box?

ReText/window.py
DanielVenturini: For now, I'm using `getOpenFileNames`

ReText/window.py
mitya57: Let's have different actions for people who want different things:

- __Formatting → image__ — to insert any image (e.g. remote one).
- __Edit → Insert images by file path__ — to insert image from local file.

I will fix it myself.

DanielVenturini:
Great. Thank you.

PR #520, opened 2020-07-16
Add surround keys

DanielVenturini:
Hi, nice app.

This pull add a functionality that surrounds a selected text with a special key rather than replace it.

e.g.:
"This text is in editor now" -> select `text` and press `(` -> "This (text) is in editor now"
"This text is in editor now" -> select `text` and press `_` -> "This \_text\_ is in editor now"
"This text is in editor now" -> select `text` and press `'` -> "This 'text' is in editor now"

It works as expected for the follow keys: `*`, `_`, `"`, `'`, `(`, `[`

### DIFF
--- a/ReText/editor.py
+++ b/ReText/editor.py
@@ -97,6 +97,16 @@ class ReTextEdit(QTextEdit):
 	orderedListPattern = re.compile("^([\\s]*)(\\d+)\\. $")
 	wordPattern = re.compile(r"\w+")
 	nonAlphaNumPattern = re.compile(r"\W")
+	surroundKeysSelfClose = [
+		Qt.Key_Underscore,
+		Qt.Key_Asterisk,
+		Qt.Key_QuoteDbl,
+		Qt.Key_Apostrophe
+	]
+	surroundKeysOtherClose = {
+		Qt.Key_ParenLeft: ')',
+		Qt.Key_BracketLeft: ']'
+	}
 
 	def __init__(self, parent):
 		QTextEdit.__init__(self)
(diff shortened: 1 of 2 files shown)


DanielVenturini:
However, other special key, grave accent ( ` ), I could not capture it as an event by Qt

mitya57:
Thanks for the PR! But what is the use case for it?

We already have <kbd>Ctrl</kbd>+<kbd>B</kbd>, <kbd>Ctrl</kbd>+<kbd>I</kbd>, <kbd>Ctrl</kbd>+<kbd>U</kbd> shortcuts that make text bold, italic and underline. We also have the Formatting menu that can turn text into a link, image or inline code.

Your PR is duplicating that functionality, isn't it?

DanielVenturini:
Yes, Retext already implements this functionality as shortcuts, but I think that my functionality is more intuitive in certain respects.  
I really agree that my functionality migth not implement the `_`, `*`, but what do you thing about the `"`, `'`, `(` and `[`?  
Specially the `(` and `"`, some others editors implements a functionality like it, that surrounds the text.

mitya57:
Ok, I don't really oppose this change, just want to understand it. Can you name editors that are also doing it?

Also maybe you can try to add a simple test for this behavior to `tests/test_editor.py`?

DanielVenturini:
Code editors, like VSCode and Sublime, and Markdown editors, like [typora](https://typora.io/), [dillinger](https://dillinger.io/) and [pandao](https://pandao.github.io/editor.md/en.html), have this functionality. In all of these markdown editors, it just works to the these keys `"`, `'`, `(` and `[`. If you select a piece of text and click one of those keys, the piece of text is surrounded.

If you think that it'd be useful to Retext, I may remove the `*` and `_` and insert some tests.

mitya57:
Thanks. No need to remove anything, it doesn't complicate code so doesn't hurt. Please try to add tests and then I will merge it.

mitya57:
Merged. Thanks again!

PR #523, opened 2020-08-06
Fix #516: Option to fully highlight wrapped-lines

### LINKED ISSUE #516: Line highlighting does not highlight full line-wrapped lines
Running 7.1.0 with Qt 5.12.8 under latest Ubuntu Focal... if I have the current line highlight option on, then with text wrapped only the one line not the several lines (i.e. line beginning to EOL) are highlighted in yellow. Commented on in https://github.com/retext-project/retext/issues/513#issuecomment-640855707

I think this option would make sense, as I want to see what is on one line and what is not, currently forcing me to turn line numbering on to see what is on one line and what is wrapped. I would make it optional, not default beaviour.

nihillum:
As stated in the title, this will resolve #516, giving the option to fully highlight the wrapped-line under the cursor, or only the single line under the cursor (current behaviour).

### DIFF
--- a/ReText/editor.py
+++ b/ReText/editor.py
@@ -373,25 +373,21 @@ class ReTextEdit(QTextEdit):
 		if globalSettings.highlightCurrentLine == 'disabled':
 			return self.setExtraSelections([])
 		elif globalSettings.highlightCurrentLine == 'cursor-line':
-			selection = QTextEdit.ExtraSelection();
+			selection = QTextEdit.ExtraSelection()
 			selection.format.setBackground(colorValues['currentLineHighlight'])
 			selection.format.setProperty(QTextFormat.FullWidthSelection, True)
 			selection.cursor = self.textCursor()
 			selection.cursor.clearSelection()
 			self.setExtraSelections([selection])
 		elif globalSettings.highlightCurrentLine == 'wrapped-line':
-			selections = [QTextEdit.ExtraSelection(),QTextEdit.ExtraSelection()]
-			selections[0].format.setBackground(colorValues['currentLineHighlight'])
-			selections[0].format.setProperty(QTextFormat.FullWidthSelection, True)
-			selections[0].cursor = self.textCursor()
-			selections[0].cursor.clearSelection()
-			selections[0].cursor.movePosition(QTextCursor.StartOfBlock)
-			selections[0].cursor.movePosition(QTextCursor.EndOfBlock, QTextCursor.KeepAnchor)
-			selections[1].format.setBackground(colorValues['currentLineHighlight'])
-			selections[1].format.setProperty(QTextFormat.FullWidthSelection, True)
-			selections[1].cursor = self.textCursor()
-			selections[1].cursor.movePosition(QTextCursor.EndOfBlock)
-			self.setExtraSelections(selections)
+			selection = QTextEdit.ExtraSelection()
+			selection.format.setBackground(colorValues['currentLineHighlight'])
+			selection.format.setProperty(QTextFormat.FullWidthSelection, True)
+			selection.cursor = self.textCursor()
+			selection.cursor.clearSelection()
+			selection.cursor.movePosition(QTextCursor.StartOfBlock)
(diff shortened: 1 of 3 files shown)


ReText/editor.py
```diff
@@ -370,14 +370,28 @@ def resizeEvent(self, event):
...
+		if globalSettings.highlightCurrentLine == 'disabled':
 			return self.setExtraSelections([])
-		selection = QTextEdit.ExtraSelection();
-		selection.format.setBackground(colorValues['currentLineHighlight'])
-		selection.format.setProperty(QTextFormat.FullWidthSelection, True)
-		selection.cursor = self.textCursor()
-		selection.cursor.clearSelection()
-		self.setExtraSelections([selection])
+		elif globalSettings.highlightCurrentLine == 'cursor-line':
+			selection = QTextEdit.ExtraSelection();
```
mitya57, line 376: code style: remove the semicolon

```suggestion
			selection = QTextEdit.ExtraSelection()
```

ReText/editor.py
```diff
@@ -370,14 +370,28 @@ def resizeEvent(self, event):
...
+			selection.format.setBackground(colorValues['currentLineHighlight'])
+			selection.format.setProperty(QTextFormat.FullWidthSelection, True)
+			selection.cursor = self.textCursor()
+			selection.cursor.clearSelection()
+			self.setExtraSelections([selection])
+		elif globalSettings.highlightCurrentLine == 'wrapped-line':
+			selections = [QTextEdit.ExtraSelection(),QTextEdit.ExtraSelection()]
```
mitya57, line 383: code style: add a space after comma

```suggestion
			selections = [QTextEdit.ExtraSelection(), QTextEdit.ExtraSelection()]
```

ReText/editor.py
```diff
@@ -370,14 +370,28 @@ def resizeEvent(self, event):
...
+			selections[0].format.setBackground(colorValues['currentLineHighlight'])
+			selections[0].format.setProperty(QTextFormat.FullWidthSelection, True)
+			selections[0].cursor = self.textCursor()
+			selections[0].cursor.clearSelection()
+			selections[0].cursor.movePosition(QTextCursor.StartOfBlock)
+			selections[0].cursor.movePosition(QTextCursor.EndOfBlock, QTextCursor.KeepAnchor)
+			selections[1].format.setBackground(colorValues['currentLineHighlight'])
+			selections[1].format.setProperty(QTextFormat.FullWidthSelection, True)
+			selections[1].cursor = self.textCursor()
+			selections[1].cursor.movePosition(QTextCursor.EndOfBlock)
```
mitya57, line 393: This second selection does not actually select any text. Why is it needed?

nihillum: It is necessary to have an highlighting that extend past the end of the last line.

nihillum: It can be removed, but doing so will make the highlighting stop at the last letter of the last line. I added it because in my opinion this way it is nicer to look at, and more consistent. Do you prefer that I remove it?

nihillum: Never mind, I just found another way that doesn't need the second selection. I will remove those lines of code.

nihillum:
All done. I was just noticing that in highlightCurrentLine() there are some repeated lines of code regarding the creation of the selection and the setting of background and property of the selection. I think the code could be refactored a bit. 
What do you think?

ReText/editor.py
mitya57: Now it doesn't work on the last line of the document :-(

mitya57:
Yes, feel free to refactor it. If it's not possible to fix the last line case without two selections, feel free to revert the second one.

ReText/editor.py
nihillum: You are right I didn't think about that... How could that be fixed?

mitya57: You can revert to your first commit. I only asked why the second selection was needed, it was not obvious :) Because it's useful, no need to remove it.

nihillum:
Done. Reverted to the precedent solution and refactored the function a bit.

mitya57:
Merged, thanks!

PR #527, opened 2020-08-30
Ordered list behavior can now be configured to repeat or increment the list item number

binokkio:
Having repeating list item numbers is useful when markdown files are in version control, without it inserting an item in the middle of the list would require changing all subsequent list item numbers and causing a bigger diff in version control. We therefore prefer to write lists like so:

```
0. Foo
0. Bar
0. Baz
0. Etcetera...
```

The changes in this PR allow the user to change the current behavior of incrementing the list item number to repeating it in the ReText preferences.

I look forward to receiving feedback and updating the PR accordingly.


mitya57:
Thanks for the very good PR! Merged.

binokkio:
That's great, thank you!

Issue #522, opened 2020-07-17
Request for feature - Side bar Folder opening option with a Project view


rohan-paul:
Absolutely great tool. 

I just felt that the following would add tremendously to its attractiveness.

If the side bar could open a full folder and I could manage files within that folder and nevigation between files from the side bar.

mitya57:
Hi @rohan-paul! I have just merged #531 which implements the side panel directory tree. Can you please check if this is what you wanted, or something else is remaining?

Issue #47, opened 2012-07-22
Directory tree side pane


mitya57:
A directory/document tree side pane like Zim's would be a fine convenience if it's not a big trouble.

Reported by: gooddead

Original Ticket: [retext/tickets/47](https://sourceforge.net/p/retext/tickets/47)

oguzhanogreden:
I started a little on a document navigation tree solution. I opted for a sidebar solution. You can see the relevant branch [here](https://github.com/oguzhanogreden/retext/tree/treenav-oo). I'm facing a decision which is likely relevant for future of this patch. Hence I'm bringing it up here.

The decision is mainly regarding the following question: _How will this scale to large documents? (= Is there a preferred large performance test file?)_

Briefly, this implementation:
A1. parses the conversion output,
A2. creates a `ReTextTreeModel` object (which inherits `QAbstractItemModel`),

When the conversion output is renewed (as signalled by `converterProcess.conversionDone`), it updates the tree by:
B1. Starts a new round of A1 & A2,
B2. Compares the new model (outcome of B1) to the present one (outcome of A2),
B3. If any difference is present (i.e. if the two lists are not identical),
B4. Starts using the new model.

Specifically, is it a reasonable idea to create trees from scratch using the converted document?

(I know that RStudio has a similar feature which gives a dropdown list of defined functions and specifically tagged comment headers. I'll check that in the meanwhile.)

Best,
Oğuzhan

oguzhanogreden:
Checking other implementations including RStudio, perhaps it's an overkill to use the QTreeView here. I can simply list the headers and give the hierarchy effect with indentations (and or by appending the header to \# or \## etc.).

This removes expand/collapse use of the hierarchy. Then again, for extremely large documents, a better Markdown approach would be switching files at one of the higher levels of hierarchy.

What do you think?

mitya57:
@oguzhanogreden I agree that a tree is overkill, just a simple list structure sounds fine to me.

I do not like the idea of converting HTML back to an ElementTree, because an ElementTree already exists during conversion (i.e. for Markdown it exists [here](https://github.com/Python-Markdown/markdown/blob/master/markdown/__init__.py#L379)).

I think it is better to write an extension that will be run after all tree processors and will save the headers structure to some attribute of Markdown instance. Then you can make PyMarkups set this structure as an attribute of [ConvertedMarkup](http://pythonhosted.org/Markups/interface.html#markups.abstract.ConvertedMarkup) object.

[...]

PR #531, opened 2020-10-09
Create side panel directory tree

### LINKED ISSUE #525: Feature: Option to show tree structured directory
Like #518, and maybe not only files, but _directories_ as well. As many mainstream editors do, you might be able to add a directory on a _side bar_. It would be great to see it and manage markdown files this way :)

xgouchet:
- A side panel is optionally added to the left of the screen, displaying a directory structure
- By default, it is hidden but can be displayed (or hidden) from the settings
- By default, when displayed, it displays the user's home folder (`~`), but the folder can be selected from the settings
- Double clicking on a file opens the file in a new tab, or make active the existing tab for this file.

This fixes #525 and partly #522


ReText/config.py
```diff
@@ -49,6 +49,26 @@ def updateButtonText(self):
...
+		            if self.fileName else '')
+		self.fileName = QFileDialog.getExistingDirectory(
+			self, self.tr('Select directory to open'), startDir)
+		self.updateButtonText()
+
+	def updateButtonText(self):
+		if self.fileName:
+			self.setText(QFileInfo(self.fileName).fileName())
+		else:
+			self.setText(self.defaultText)
```
mitya57, line 71: The `__init__` and `updateButtonText` methods of this class are copied from `FileSelectButton` without any changes. Maybe you can split them out into a base class, to avoid duplication?

ReText/__init__.py
```diff
@@ -71,6 +71,8 @@ def getBundledIcon(iconName):
 	'defaultMarkup': markups.MarkdownMarkup.name,
 	'defaultPreviewState': 'editor',
 	'detectEncoding': True,
+	'directoryPath': expanduser("~"),
+	'directoryTree': False,
```
mitya57, line 75: Maybe this option should be named `showDirectoryTree` to make it more obvious that it's a true/false option?

Also please document the new options in `configuration.md`.

ReText/config.py
```diff
@@ -229,6 +253,8 @@ def saveSettings(self):
 				value = configurator.currentData()
 			elif isinstance(configurator, FileSelectButton):
 				value = configurator.fileName
+			elif isinstance(configurator, DirectorySelectButton):
+				value = configurator.fileName
```
mitya57, line 257: If you make a base class like I suggested above, then it will be possible to list only that class here, instead of two classes.

ReText/window.py
```diff
@@ -46,7 +47,8 @@
  QTextDocument, QTextDocumentWriter
 from PyQt5.QtWidgets import QAction, QActionGroup, QApplication, QCheckBox, \
  QComboBox, QDesktopWidget, QDialog, QFileDialog, QFontDialog, QInputDialog, \
- QLineEdit, QMainWindow, QMenu, QMessageBox, QTabWidget, QToolBar
+ QLineEdit, QMainWindow, QMenu, QMessageBox, QTabWidget, QToolBar, QSplitter, \
+ QTreeView, QFileSystemModel
```
mitya57, line 51: Please keep this list sorted.

ReText/window.py
```diff
@@ -428,6 +434,26 @@ def dropEvent(e):
 		self.tabWidget.dropEvent = dropEvent
 		self.tabWidget.setTabBarAutoHide(globalSettings.tabBarAutoHide)
 
+	def initDirectoryTree(self, visible, path):
+		if visible:
+			self.fileSystemModel = QFileSystemModel(self.treeView)
+			self.fileSystemModel.setRootPath(path)
```
mitya57, line 440: Maybe we should show only supported files (like in `openFile` method), not all files?

ReText/window.py
```diff
@@ -428,6 +434,26 @@ def dropEvent(e):
...
+			self.treeView.setModel(self.fileSystemModel)
+			self.treeView.setRootIndex(self.fileSystemModel.index(path))
+			self.treeView.setColumnHidden(1, True)
+			self.treeView.setColumnHidden(2, True)
+			self.treeView.setColumnHidden(3, True)
+			self.treeView.setHeaderHidden(True)
+			self.splitter.setSizes([self.width() / 5, self.width() * 4 / 5])
+		else:
+			self.splitter.setSizes([0, self.width()])
```
mitya57, line 449: This way the splitter is still visible, and it is possible to drag it to the right — which results in a blank space.

I think it's better to call `set.treeView.setVisible(False)` in this case — like we do for the tab's QSplitter when live preview is disabled.

mitya57 (changes requested):
Thanks for this PR — this is one of the most requested features for ReText (the earliest opened issue #47 is exactly about this) :+1:

I have left some comments though.

xgouchet:
Thanks for the review and the comments :bow:

mitya57:
Good work! Thank you again! :rocket:

PR #556, opened 2021-04-17
Turkish translation was created

serkan-maker:
I finished the translation in Transifex. Can you check it?

### DIFF
--- a/locale/retext_tr.ts
+++ b/locale/retext_tr.ts
@@ -244,7 +244,7 @@
     <message>
         <location filename="../ReText/fakevimeditor.py" line="61"/>
         <source>No write since last change (add ! to override)</source>
-        <translation>Son değişiklikten bu yana yazma yok (geçersiz kılmak için! Ekleyin)</translation>
+        <translation>Son değişiklikten bu yana yazma yok (geçersiz kılmak için ekleyin !)</translation>
     </message>
     <message>
         <location filename="../ReText/fakevimeditor.py" line="118"/>
@@ -292,9 +292,9 @@
         <source>The file &apos;%s&apos; does not exist.
 
 Do you want to create it?</source>
-        <translation>&apos;% S&apos; dosyası mevcut değil.
+        <translation>&apos;%s&apos; dosyası mevcut değil.
 
-Onu yaratmak ister misin?</translation>
+Onu oluşturmak ister misin?</translation>
     </message>
     <message>
         <location filename="../ReText/tab.py" line="515"/>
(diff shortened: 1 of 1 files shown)


locale/retext_tr.ts
```diff
@@ -0,0 +1,849 @@
...
+        <location filename="../ReText/tab.py" line="482"/>
+        <source>Create missing file?</source>
+        <translation>Eksik dosya oluşturulsun mu?</translation>
+    </message>
+    <message>
+        <location filename="../ReText/tab.py" line="482"/>
+        <source>The file &apos;%s&apos; does not exist.
+
+Do you want to create it?</source>
+        <translation>&apos;% S&apos; dosyası mevcut değil.
```
mitya57, line 295:
```suggestion
        <translation>&apos;%s&apos; dosyası mevcut değil.
```

locale/retext_tr.ts
```diff
@@ -0,0 +1,849 @@
...
+    </message>
+    <message>
+        <location filename="../ReText/window.py" line="1289"/>
+        <source>Select one or several images to open</source>
+        <translation>Açmak için bir veya birkaç resim seçin</translation>
+    </message>
+    <message>
+        <location filename="../ReText/window.py" line="1307"/>
+        <source>ReText %s (using PyMarkups %s)</source>
+        <translation>ReText %s (using PyMarkups %s)</translation>
```
mitya57, line 794: Here you can translate the word ‘using’.

locale/retext_tr.ts
```diff
@@ -0,0 +1,849 @@
...
+        <source>Set as default</source>
+        <translation>Öntanımlı olarak ata</translation>
+    </message>
+</context>
+<context>
+    <name>Proxy</name>
+    <message>
+        <location filename="../ReText/fakevimeditor.py" line="61"/>
+        <source>No write since last change (add ! to override)</source>
+        <translation>Son değişiklikten bu yana yazma yok (geçersiz kılmak için! Ekleyin)</translation>
```
mitya57, line 247: Here the string means that one should literally type an exclamation mark. Did you translate it correctly? It is suspicious to me that `Ekleyin` is capitalized. And maybe add a space before `!`?

mitya57 (changes requested):
Thank you very much for your translation!

I usually grab translations from Transifex before every release (or once in a few months), but as you submitted a PR, I will review it here. Left some comments.

serkan-maker:
Thank you for this program. yes you are right the translation is wrong. i applied the fix, now i install it.
[retext_tr.ts.zip](https://github.com/retext-project/retext/files/6331974/retext_tr.ts.zip)

mitya57 (approved):
Thank you! Now it looks good, I will merge it a bit later.

mitya57:
Merged, thanks again. Now I am looking at the translation in desktop file.

It currently says:

> Markdown ve Yeniden Yapılandırılmış Metin için basit ama güçlü düzenleyici

Do you think reStructuredText should be translated? It is a proper name, not two words. See <https://en.wikipedia.org/wiki/ReStructuredText>.

So maybe better to make it “Markdown ve reStructuredText için basit ama güçlü düzenleyici”?

Or just “Markdown ve reStructuredText için basit düzenleyici”, as “but powerful” was not part of the original string here.

serkan-maker:
I corrected in transifex

mitya57:
Applied in b39ad26a32a4930b. Thanks.

PR #592, opened 2022-08-06
Move some of the config options to a separate cache file

okkomakkonen:
Some of the settings described in the `configuration.md` file are said to be editable by ReText only, not by the user. Therefore, it makes sense to store them in separate files.

One possible scenario would be if the user wanted to save their configuration. Now they can save the `ReText.conf` file and use it on another instance of ReText without affecting the config entries that should not be user editable. I came across this issue when I wanted to save the `ReText.conf` file to be tracked by my dotfiles repository, but I had trouble with the `recentFileList` entry, since I did not want to track those.

There are still some issues with this implementation. One of them is that the cache file is created with the same QSettings class as the main config file and its name does not perfectly match its intention.

### DIFF
--- a/ReText/__init__.py
+++ b/ReText/__init__.py
@@ -32,7 +32,7 @@ if not str(settings.fileName()).endswith('.conf'):
 	settings = QSettings(QSettings.Format.IniFormat, QSettings.Scope.UserScope,
 		'ReText project', 'ReText')
 
-cache = QSettings('ReText project', 'ReText cache')
+cache = QSettings('ReText project', 'cache')
 
 if not str(cache.fileName()).endswith('.conf'):
 	# We are on Windows probably
--- a/ReText/__init__.py
+++ b/ReText/__init__.py
@@ -138,17 +138,30 @@ def getSettingsFilePath():
 
 
 class ReTextSettings:
-	def __init__(self):
-		for option in configOptions:
-			value = configOptions[option]
-			object.__setattr__(self, option, readFromSettings(
-				option, type(value), default=value, settings=settings))
+
+	def __init__(self, settings, defaults):
+		# We have to do this to go around the custom __setattr__ method
+		object.__setattr__(self, "settings", settings)
+		object.__setattr__(self, "defaults", defaults)
+		for option in defaults:
+			default = defaults[option]
+			if isinstance(default, list):
+				object.__setattr__(self, option, readListFromSettings(
+					option, settings=settings))
+			else:
+				object.__setattr__(self, option, readFromSettings(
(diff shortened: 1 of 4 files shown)


ReText/__init__.py
```diff
@@ -32,6 +32,13 @@
 	settings = QSettings(QSettings.Format.IniFormat, QSettings.Scope.UserScope,
 		'ReText project', 'ReText')
 
+cache = QSettings('ReText project', 'ReText cache')
```
mitya57, line 35: Let's name it just `cache.conf`, to make it more friendly to people who open files from terminal.

```suggestion
cache = QSettings('ReText project', 'cache')
```

ReText/__init__.py
```diff
@@ -32,6 +32,13 @@
...
+
+if not str(cache.fileName()).endswith('.conf'):
+	# We are on Windows probably
+	cache = QSettings(QSettings.Format.IniFormat, QSettings.Scope.UserScope,
+		'ReText project', 'ReText cache')
```
mitya57, line 40:
```suggestion
		'ReText project', 'cache')
```

ReText/__init__.py
```diff
@@ -150,6 +162,44 @@ def getEditorFont(self):
 			font.fromString(self.editorFont)
 		return font
 
+class ReTextCache:
```
mitya57, line 165: Instead of introducing another class, what do you think about making a generic class and passing `settings` and `options` to its constructor?

tests/test_window.py
```diff
@@ -57,8 +57,6 @@ class TestWindow(unittest.TestCase):
 
     def setUp(self):
         warnings.simplefilter("ignore", Warning)
-        self.readListFromSettingsMock = patch('ReText.window.readListFromSettings', return_value=[]).start()
-        self.writeListToSettingsMock  = patch('ReText.window.writeListToSettings').start()
```
mitya57, line 61: Now maybe you should add a mock for `globalCache`?

The purpose of these mocks is to prevent tests from messing up with user's configuration (and to make them pass when configuration is non-standard).

mitya57 (changes requested):
Thanks for your PR!

ReText/__init__.py
okkomakkonen: Yes, I think this is the logical thing to do!

tests/test_window.py
okkomakkonen: I added the mock for `globalCache`.

mitya57:
Merged, thank you! :tada:

PR #634, opened 2023-10-04
Show link on hover

red-kite:
Please:
- check that I'm doing right with `super().__init__(parent)`
- verify and change positions of inserted code
- eventually `.setText('')` additional to `setVisible(False)`
- maybe test it :-)
- …?

Thank you for creating and maintaining ReText!

### DIFF
--- a/ReText/dialogs.py
+++ b/ReText/dialogs.py
@@ -79,3 +79,28 @@ class EncodingDialog(QDialog):
 			button.setEnabled(False)
 		else:
 			button.setEnabled(True)
+
+
+class urlPopup(QLabel):
+
+	def __init__(self, parent):
+		super().__init__(parent)
+		self.p = parent
+
+		self.setStyleSheet('''
+			border: 1px solid rgba(50, 50, 50, 100);
+			border-radius: 3px;
+			background: rgba(250, 250, 250, 250);
+		''')
+		self.fh = self.p.fontMetrics().height()
+		self.setVisible(False)
+
+	def popup(self, url):
+		if url:
+			self.setText(url)
+			bot = self.p.rect().bottom()
+			tw = self.p.fontMetrics().horizontalAdvance(url)
+			self.setGeometry(-2, bot-self.fh-5, tw+10, self.fh+10)
+			self.setVisible(True)
+		else:
+			self.setVisible(False)
--- a/ReText/dialogs.py
+++ b/ReText/dialogs.py
@@ -79,28 +79,3 @@ class EncodingDialog(QDialog):
 			button.setEnabled(False)
 		else:
 			button.setEnabled(True)
-
-
-class urlPopup(QLabel):
-
-	def __init__(self, parent):
-		super().__init__(parent)
-		self.p = parent
-
-		self.setStyleSheet('''
-			border: 1px solid #64323232;
-			border-radius: 3px;
-			background: #FAFAFAFA;
-		''')
-		self.fh = self.p.fontMetrics().height()
-		self.setVisible(False)
-
-	def popup(self, url):
-		if url:
-			self.setText(url)
-			bot = self.p.rect().bottom()
-			tw = self.p.fontMetrics().horizontalAdvance(url)
-			self.setGeometry(-2, bot-self.fh-5, tw+10, self.fh+10)
-			self.setVisible(True)
-		else:
-			self.setVisible(False)
--- a/ReText/dialogs.py
+++ b/ReText/dialogs.py
@@ -88,9 +88,9 @@ class urlPopup(QLabel):
 		self.p = parent
 
 		self.setStyleSheet('''
-			border: 1px solid rgba(50, 50, 50, 100);
+			border: 1px solid #64323232;
 			border-radius: 3px;
-			background: rgba(250, 250, 250, 250);
+			background: #FAFAFAFA;
 		''')
 		self.fh = self.p.fontMetrics().height()
 		self.setVisible(False)
--- a/ReText/webenginepreview.py
+++ b/ReText/webenginepreview.py
@@ -43,6 +43,7 @@ class ReTextWebEnginePage(QWebEnginePage):
         self.tab = tab
         self.interceptor = ReTextWebEngineUrlRequestInterceptor(self)
         self.setUrlRequestInterceptor(self.interceptor)
+        self.linkHovered.connect(self.onLinkHover)
 
     def setScrollPosition(self, pos):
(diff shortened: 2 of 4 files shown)


ReText/dialogs.py
```diff
@@ -79,3 +79,28 @@ def handleTextChanged(self, value):
 			button.setEnabled(False)
 		else:
 			button.setEnabled(True)
+
+
+class urlPopup(QLabel):
```
mitya57, line 84: Please move this class to `webenginepreview.py`, since it's not a dialog, but part of the previewer. 

Also please capitalize the first letter: `UrlPopup`.

ReText/dialogs.py
```diff
@@ -79,3 +79,28 @@ def handleTextChanged(self, value):
...
+			background: #FAFAFAFA;
+		''')
+		self.fh = self.p.fontMetrics().height()
+		self.setVisible(False)
+
+	def popup(self, url):
+		if url:
+			self.setText(url)
+			bot = self.p.rect().bottom()
+			tw = self.p.fontMetrics().horizontalAdvance(url)
```
mitya57, line 102: Please use longer names for the variables to make them more meaningful: `self.fontHeight`, `bottom`, `textWidth`.

ReText/window.py
```diff
@@ -95,6 +95,7 @@ def __init__(self, parent=None):
 		self.addToolBar(Qt.ToolBarArea.BottomToolBarArea, self.searchBar)
 		self.toolBar.setVisible(not globalSettings.hideToolBar)
 		self.editBar.setVisible(not globalSettings.hideToolBar)
+		self.urlPopup = urlPopup(self)
```
mitya57, line 98: Please remove it from here and make it an attribute of ReTextWebEnginePage.

mitya57 (changes requested):

ReText/window.py
red-kite: Good idea, nicely keeping it modular. Before, I thought it would be sufficient to have one instance of UrlPopup for one Retext window.

Latest implementation explicitly lets it stick to 'window': `def __init__(self, window):` (…and still has to (?))

red-kite: Off-topic / note by the way:

Meanwhile I had:
``` python
class UrlPopup(QLabel):
    def __init__(self, parent):
        super().__init__(parent.tab.p)
        self.page = parent
    def pop(…):
        windowBottom = self.page.tab.rect().bottom()
```
(only relevant lines, tab also has .rect())
--> Doing so I realized that _tab's bottom is well above the window's_ bottom! (But may be associated with various coordinate-systems window/tab)

mitya57:
It does not look nice with dark themes:
Screenshot

I recommend that you take colors from QPalette where possible, rather than hard-coding them. If QPalette doesn't have the needed colors, please define both light and dark versions, you can add them after this line:
https://github.com/retext-project/retext/blob/adeba3840c434a0e59b86fc584161820cb94714e/ReText/editor.py#L54

ReText/window.py
mitya57: Tab's bottom is calculated from tab's top, and window's bottom from window's top. And there is menu bar, toolbar and tabs panel between window's top and tab's top. So they make the difference that you see.

ReText/webenginepreview.py
```diff
@@ -37,12 +38,40 @@ def interceptRequest(self, info):
...
+
+
 class ReTextWebEnginePage(QWebEnginePage):
     def __init__(self, parent, tab):
         QWebEnginePage.__init__(self, parent)
         self.tab = tab
         self.interceptor = ReTextWebEngineUrlRequestInterceptor(self)
         self.setUrlRequestInterceptor(self.interceptor)
+        self.urlPopup = UrlPopup(self.tab.p)
+        self.linkHovered.connect(self.onLinkHover)
```
mitya57, line 74: I think you can do this:
```suggestion
        self.linkHovered.connect(self.urlPopup.pop)
```

and remove the `onLinkHover` method.

mitya57 (changes requested):

mitya57:
So the only thing remaining is my comment about dark themes.

red-kite:
Please excuse being that quiet on this for so long.
Thank you @mitya57 for guiding towards the solution via QPalette *or* hardcoded colors. I'd prefer latter for decent appearance with soft border and very little transparency (finetuning possible)
…but ran into issues:

* Figure out how to access editor's colors-dictionary (Use example in highlighter.py: `from ReText.editor import getColor`)
* Use QColor in stylesheet of UrlPopup ('QColor' has no attribute 'HexArgb') -- Got a solution, a more elegant might exist.
* Major: **I can't test dark-mode** (because I don't use such). I tried (again) in a gnome shell but only window decoration gets dark (stylesheet disabled)


Thanks to motivation by @donjan I tried some more and pushed https://github.com/retext-project/retext/pull/634/commits/8935a93595c6701c8b90e74eca6f3a0e4b86a012 for further use…

-> Help appreciated

----
Some alternative code (palette usage found on the internet last year, *no alpha*):

```
        self.setStyleSheet('''
            border: 1px solid palette(mid);
            border-radius: 3px;
            background: palette(base);
        ''')
```
Resources:

https://stackoverflow.com/a/2756376 suggests ("not to use the palette and go for Qt Style Sheet")
https://stackoverflow.com/a/6876509 (long struggle to convert QColor to HexArgb!)

ReText/webenginepreview.py
```diff
@@ -37,12 +39,52 @@ def interceptRequest(self, info):
             info.block(True)
 
 
+class UrlPopup(QLabel):
+    def __init__(self, window):
+        super().__init__(window)
+        self.window = window
+
+        def hex_rgba(color):
+            ''' color:QColor '''
```
mitya57, line 48: Can this be made a static method or a top-level function?
This way we will have only one instance of this function for all labels.

Also, you can import QColor and use native type annotation (`color: QColor`).

mitya57 (changes requested):
Thank you very much. Can I ask you for a couple more things?

red-kite:
see commit msg, *note that dark-mode still not tested by me*

unsure about position of hex_rgba() in file -- kept it close to it's class

(to-string-conversion of QColor could be part of getColor(colorName, rgba=False) but I don't like this because function gets blown up for other use)

mitya57 (approved):

mitya57:
Thank you!

Tested with dark theme (QT_STYLE_OVERRIDE=Adwaita-Dark), it looks good!

Screenshot with Adwaita-Dark

Although, the transparency is barely visible, so maybe it's an overcomplication and using the palette color would be enough. But as you already implemented it this way, I did not change.

PR #658, opened 2025-03-09
Add alternative toolbar for Formatting options

sendevent:
This PR introduces an alternative way to display the list of available **Formatting** options in the toolbar. Instead of using a combobox, users can now enable a dedicated toolbar for one-click access. The default combobox is displayed when the new toolbar is disabled via the **Configuration** dialog or the toolbars' context menu.

Additionally, this PR includes:
- Small refactorings in `window.py` to improve the code structure.
- Adjustments to the **Config** dialog, specifically in the **Interface** tab, to enhance organization.

retext_1200_20


ReText/__init__.py
```diff
@@ -59,6 +59,9 @@ def getBundledIcon(iconName):
     'font': '',
     'handleWebLinks': False,
     'hideToolBar': False,
+    'showToolBarFile': True,
+    'showToolBarEdit': True,
+    'showToolBarFormat': True,
```
mitya57, line 64: Maybe these names would sound more naturally in English?

```suggestion
    'showFileToolBar': True,
    'showEditToolBar': True,
    'showFormatToolBar': True,
```

ReText/config.py
```diff
@@ -33,7 +33,7 @@
     QSpinBox,
     QTabWidget,
     QVBoxLayout,
-    QWidget,
+    QWidget, QGroupBox, QHBoxLayout,
```
mitya57, line 36: Please use one import per line and keep the list sorted (you can use `isort` tool for that).

ReText/config.py
```diff
@@ -289,8 +321,13 @@ def applySettings(self):
...
         self.parent.tabWidget.setTabBarAutoHide(globalSettings.tabBarAutoHide)
-        self.parent.toolBar.setVisible(not globalSettings.hideToolBar)
-        self.parent.editBar.setVisible(not globalSettings.hideToolBar)
+
+        self.parent.manage_toolbars_visibility()
+        self.configurators["showToolBarFile"].setChecked(self.parent.toolBar.toggleViewAction().isChecked())
+        self.configurators["showToolBarEdit"].setChecked(self.parent.editBar.toggleViewAction().isChecked())
+        self.configurators["showToolBarFormat"].setChecked(self.parent.formatBar.toggleViewAction().isChecked())
+
+
```
mitya57, line 330: Please don't use two empty lines inside a function, remove one.

ReText/retext.qrc
```diff
@@ -0,0 +1,28 @@
+<RCC>
+    <qresource prefix="/icn_format">
+        <file>icons/format/light/blockquote.svg</file>
+        <file>icons/format/light/bold.svg</file>
```
mitya57, line 4: At least for bold, italics and underline (and probably also for images, bullets and numbering), it is possible to use theme icons:
- https://specifications.freedesktop.org/icon-naming-spec/latest/
- https://github.com/KDE/oxygen-icons/tree/master/22x22/actions

If the icons exist in the user theme, then it’s preferable to use those, otherwise we use the built-in ones.

But for the built-in ones, please use the existing mechanism (`getBundledIcon` function) instead of inventing a new one. This way you won't need the ugly `retext_rc.py` file.

ReText/window.py
```diff
@@ -38,6 +38,8 @@
     ReTextWebEnginePreview,
 )
 from ReText.tabledialog import InsertTableDialog
+from ReText.editor import detectThemeVariant
+from ReText import retext_rc
```
mitya57, line 42: Please add this to line 29 instead.

ReText/window.py
```diff
@@ -308,44 +325,146 @@ def __init__(self, parent=None):
...
-            trig=lambda: self.insertFormatting('italic'))
+            trig=self._on_formatting_action)
         self.actionUnderline = self.act(self.tr('Underline'), shct=QKeySequence.StandardKey.Underline,
-            trig=lambda: self.insertFormatting('underline'))
+            trig=self._on_formatting_action)
+
+    def _create_menus(self):
+        self._create_menu_file()
+        self._create_menu_edit()
+        self._create_menu_help()
```
mitya57, line 340: Similarly, maybe better to name `createFileMenu`, `createEditMenu`, `createHelpMenu`, etc?

By the way, this file uses `camelCase` for consistency with Qt. Also, there is no need for leading underscore, since effectively everything is private.

Also, I would be extra grateful if you submit this refactoring as a separate PR, this way I can merge it sooner.

mitya57 (changes requested):
Sorry for delay with reviewing this. I have left some comments now.

sendevent:
> Sorry for delay with reviewing this. I have left some comments now.

Sure thing, I'll handle all these in a few days, probably this weekend. 

Regarding the code formatting, do you use any automation tools so you could share its settings?

mitya57:
> Regarding the code formatting, do you use any automation tools so you could share its settings?

In new code I use `black` or `ruff format` (which use similar style), but ReText is very old code that doesn't use any best practices for formatting (it even had tabs until recently). And reformatting it would make `git blame` unusable.

Adding new `black`-style code is good.

PR #674, opened 2025-10-05
Reverse scroll (scrolling the preview will also scroll source)

carli2:
i want to scroll in preview, too and it will synchronize back the scroll position to source code view

### DIFF
--- a/ReText/webenginepreview.py
+++ b/ReText/webenginepreview.py
@@ -176,8 +176,7 @@ class ReTextWebEnginePreview(QWebEngineView):
         self.editBox.resized.connect(self._handleEditorResized)
 
         # When preview is scrolled, update the editor scroll accordingly
-        if hasattr(webPage, 'scrollPositionChanged'):
-            webPage.scrollPositionChanged.connect(self.syncscroll.handlePreviewScrolled)
+        webPage.scrollPositionChanged.connect(self.syncscroll.handlePreviewScrolled)
 
         # Scroll the preview when the mouse wheel is used to scroll
         # beyond the beginning/end of the editor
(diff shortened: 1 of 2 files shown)


ReText/webenginepreview.py
```diff
@@ -172,6 +175,10 @@ def __init__(self, tab,
         self.editBox.verticalScrollBar().valueChanged.connect(self.syncscroll.handleEditorScrolled)
         self.editBox.resized.connect(self._handleEditorResized)
 
+        # When preview is scrolled, update the editor scroll accordingly
+        if hasattr(webPage, 'scrollPositionChanged'):
```
mitya57, line 179: This signal was introduced in Qt 5.7, and we require Qt 6. Maybe the `if` is not needed?

And the same below on line 236.

mitya57:
Sorry for the delay with reviewing this (I was one week out, and another week keeping up with backlog).

I tested this with various files, and it mostly looks good. However, on this file I see an interesting behavior:
<https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docs/ref/doctree.rst?format=raw>

When you open live preview and try to scroll the editor panel, it gets stuck around line 5663: attempt to scroll below this line results in both editor and preview being reverted back to the same position.

Maybe it's because the posmap sequence goes down at one point:
```html
<p data-posmap="5656">Entity definition:</p>
<pre class="literal-block" data-posmap="5678">(#PCDATA | <a class="reference internal" href="#inline-elements-1">%inline.elements</a>;)*</pre>
<p data-posmap="5662">The <span class="docutils literal">%text.model</span> parameter entity is directly employed in the content
models of the following elements: <a class="reference internal" href="#abbreviation">&lt;abbreviation&gt;</a>,
```

Any chance you can do something about this?

P.S. if you changed operator `/` to `//`, there is no more need to wrap into `int()`, the result is already int.

carli2:
I found a fix that now also handles your big document really nicely.

carli2:
I hope, this commit fixes it. I tested with doctree.rst and I was not able to observe similar flickering events.

PR #685, opened 2026-05-12
Add copy file path action

Symphonic-eMotions:
A feature to copy the active document path to the clipboard. Also added tests for this feature


mitya57:
Thank you for the pull request!

I think this should be in the File menu:
- First, the Edit menu is already too long (I'm even thinking about splitting it into two menus).
- Second, the Cut/Copy/Paste actions go together and are related to the editor. This action does something different.

Symphonic-eMotions:
Thank you for your reply. I've moved the option to the File menu as you suggested and ran all tests, including the new test_copyFilePathCopiesActiveDocumentPath, successfully.

mitya57:
Merged, thank you!

Issue #137, opened 2015-02-18
fileChanged is only triggered once.


mitya57:
if a markdown file is changed outside of retext, the fileChanged function is only triggered once -- the first time the file changed outside of retext.

Subsequent changes outside retext do not trigger further calls to the fileChanged function.

This was observed in the latest retext as from git (2015-02-18).

Reported by: joze

Original Ticket: [retext/tickets/137](https://sourceforge.net/p/retext/tickets/137)

mitya57:
Closing because of no feedback.

PR #691, opened 2026-09-08
Retry re-watching a file after it is replaced instead of edited in place

unbroken75:
Refs #137.

## Problem

Auto-reload of externally modified files stops working after the first change
when the other application does not write the file in place.

Most editors and many tools save with the "atomic save" pattern: write the new
content to a temporary file, then rename it over the original. The path then
refers to a different file than the one the OS-level watch was created for, so
`QFileSystemWatcher` reports the first change and then goes silent for that
path — the symptom described in #137 ("fileChanged is only triggered once").

`fileChanged()` already worked around this by re-adding the path after every
change, but it called `addPath()` once and ignored its return value. If the new
file was not in place at that exact moment — a save that deletes and recreates
the file, or simply a slow filesystem — `addPath()` failed and the file stayed
unwatched for the rest of the session, with no way for the user to notice or
recover other than reopening the tab.

## Change

`rewatchFile()` replaces the single blind `addPath()` call:

- it skips files that are already watched;
- it checks that `addPath()` actually succeeded;
- if it did not, it retries every 200 ms for up to 10 attempts (~2 s), giving
up early if the tab is closed in the meantime.

It is also used when opening a file, since the same race exists if the file
happens to be written at the moment it is opened.

## Tests

Two tests are added to `tests/test_window.py`:

- `test_reloadFileReplacedAtomically` — a file replaced twice via
`os.replace()`; both changes have to be picked up, which requires the watch
to be re-created after the first one.
- `test_rewatchFileAfterItReappears` — the file is removed, the watcher is
given time to notice, and the file is then recreated. This is the case the
retry exists for.

Both follow the existing convention of the sibling watcher tests and are
skipped on Windows, so they run on the Linux jobs. I verified them on Debian:
the whole suite passes, and `test_rewatchFileAfterItReappears` fails without
this change — checked by setting `MAX_REWATCH_ATTEMPTS` to 0, which reduces
`rewatchFile()` to the previous single `addPath()` call, and the watcher is
then left with no watched files at all. So it covers the failure mode from the
issue, not just the atomic-rename case.

Out of curiosity I also ran all four watcher tests unskipped on Windows 10 with
PyQt6 6.11 and they pass there, once the path assertion is adjusted to native
separators — `QFileSystemWatcher.files()` appears to keep backslashes on
Windows, so the existing `fileName.replace('\\', '/')` in those tests looks
like it would not match there. I left that alone to keep this change focused.

I also verified the fix by hand in the running application: with an external
script replacing the open file every few seconds, the preview now follows every
change, where before it updated only once.

### DIFF
--- a/tests/test_window.py
+++ b/tests/test_window.py
@@ -33,7 +33,7 @@ from PyQt6.QtTest import QTest
 from PyQt6.QtWidgets import QApplication, QMessageBox
 
 import ReText
-from ReText.window import ReTextWindow
+from ReText.window import MAX_REWATCH_ATTEMPTS, REWATCH_RETRY_INTERVAL, ReTextWindow
 
 path_to_testdata = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'testdata')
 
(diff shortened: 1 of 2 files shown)


tests/test_window.py
```diff
@@ -520,5 +520,85 @@ def test_reloadFileModified(self, messageBoxExecMock, messageBoxWarningMock):
...
+        self.assertEqual(window.fileSystemWatcher.files(), [fileName.replace('\\', '/')])
+        editBox = window.currentTab.editBox
+        self.assertEqual(editBox.toPlainText(), 'first content')
+        app.processEvents()
+
+        os.remove(fileName)
+        QTest.qWait(100)  # let the watcher notice that the file is gone
+        with open(fileName, 'w', encoding='utf-8') as tempFile:
+            tempFile.write('recreated externally')
+        QTest.qWait(10 * REWATCH_RETRY_INTERVAL)
```
mitya57: 10 is `MAX_REWATCH_ATTEMPTS`? Can we import that constant instead of duplicating it here?

mitya57 (approved):
This mostly looks good, just one minor comment.

tests/test_window.py
unbroken75: Right, that was MAX_REWATCH_ATTEMPTS — fixed now.

mitya57:
Merged, thank you!

PR #692, opened 2026-09-09
Add a savePreviewState option to restore the preview state per document

unbroken75:
Add savePreviewState, which stores the preview state of each of the last opened documents when ReText exits and restores it when those documents are reopened. It builds on the existing session restore, so it goes together with openLastFilesOnStartup, and it falls back to defaultPreviewState for documents with no saved state.

The state is kept in the cache file next to lastFileList, using the same names defaultPreviewState already accepts.

The tests need one fix to the existing setUp to be able to cover this: the settings mock is built from configOptions, which only holds option values, so it has no working getPreviewFont(). Any tab left in a preview state then passes a MagicMock to QTextDocument.setDefaultFont, and since that happens in a slot, the exception aborts the whole test process instead of failing a single test.

### DIFF
--- a/ReText/window.py
+++ b/ReText/window.py
@@ -102,6 +102,8 @@ previewStatesByName = {
     'live-preview': PreviewLive,
 }
 
+previewStateNames = {state: name for name, state in previewStatesByName.items()}
+
 # See rewatchFile() for what these are used for.
 REWATCH_RETRY_INTERVAL = 200  # milliseconds
 MAX_REWATCH_ATTEMPTS = 10
@@ -494,8 +496,15 @@ class ReTextWindow(QMainWindow):
         self.fileSystemWatcher.fileChanged.connect(self.fileChanged)
 
     def restoreLastOpenedFiles(self):
-        for file in globalCache.lastFileList:
-            self.openFileWrapper(file)
+        previewStates = []
+        if globalSettings.savePreviewState:
+            previewStates = globalCache.lastPreviewStateList
+        for index, file in enumerate(globalCache.lastFileList):
+            # The saved state wins over defaultPreviewState.
+            previewState = None
+            if index < len(previewStates):
+                previewState = previewStatesByName.get(previewStates[index], PreviewDisabled)
+            self.openFileWrapper(file, previewState)
 
         # Show the tab of last opened file
         lastTabIndex = globalCache.lastTabIndex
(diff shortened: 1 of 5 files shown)


ReText/window.py
```diff
@@ -494,8 +496,18 @@ def __init__(self, parent=None):
...
+            previewStates = globalCache.lastPreviewStateList
+        for index, file in enumerate(globalCache.lastFileList):
+            if not file:
+                continue
             self.openFileWrapper(file)
+            if index < len(previewStates):
+                # The saved state wins over defaultPreviewState, which
+                # openFileWrapper has just applied.
+                self.setPreviewState(
+                    previewStatesByName.get(previewStates[index], PreviewDisabled))
```
mitya57: Maybe add an additional argument to `openFileWrapper()` and `createTab()` so they set the desired preview state directly, rather than updating it afterwards?

tests/test_window.py
```diff
@@ -80,6 +86,14 @@ def tearDown(self):
     # Helper functions
     #
 
+    @staticmethod
+    def create_temp_document():
```
mitya57: I think this function is not needed. All places that call it could just use NamedTemporaryFile context manager (maybe with `delete_on_close=False` if you need that), which would delete the file automatically when you leave the context.

tests/test_window.py
```diff
@@ -600,5 +614,66 @@ def test_rewatchFileAfterItReappears(self, messageBoxExecMock, messageBoxWarning
...
+    def test_savePreviewState(self):
+        self.globalSettingsMock.openLastFilesOnStartup = True
+        self.globalSettingsMock.savePreviewState = True
+        fileName = self.create_temp_document()
+
+        window = ReTextWindow()
+        window.openFileWrapper(fileName)
+        window.setPreviewState(PreviewLive)
+        self.assertEqual(window.currentTab.previewState, PreviewLive)
+        window.closeEvent(QCloseEvent())
```
mitya57: You can just use `window.close()` and remove QCloseEvent import.

ReText/window.py
unbroken75: Good catch, done.

tests/test_window.py
unbroken75: The helper is gone. I could not use NamedTemporaryFile for it though: delete_on_close needs Python 3.12 while the project supports 3.9, and without it the handle stays open, which makes reopening the path fail on Windows — these tests are not skipped there and CI runs the py39-windows job. TemporaryDirectory fails on Windows too, with a "file used by another process" error on cleanup, because openFileWrapper() ends up in QDir.setCurrent() on the document's directory, so the process' working directory is inside the temporary one.
These tests only need an existing readable file, so they now use path_to_testdata/existing_file.md like most of the other tests in this file. No temporary file, no cleanup at all.

tests/test_window.py
unbroken75: Done

mitya57 (approved):

mitya57:
Merged, thank you!

